PR #19, opened 2022-01-25
profile: fix m1 support

christophe-lunarg:


richard-lunarg:
This did fix the failure when calling vpGetPhysicalDeviceProfileSupport on the M1 mac.

PR #25, opened 2022-01-26
profiles layer: Use append_if macro to make code more readable

ziga-lunarg:


christophe-lunarg (approved):
MACROs are not my cup of tea but that's up to you!

mark-lunarg:
Yikes, is there any other way to do this?  Macros are usually against the Lunarg styleguide.

ziga-lunarg:
I guess I can change this to a function in a different PR.

PR #32, opened 2022-01-26
profiles layer: Update documentation and set default values

ziga-lunarg:


layer/TUTORIAL.md
```diff
@@ -110,17 +110,17 @@ Enables emulation of the `VK_KHR_portability_subset` extension.
 - Environment Variable: `VK_KHRONOS_PROFILES_SIMULATE_CAPABILITIES`
 - `vk_layer_settings.txt` Option: `khronos_profiles.simulate_capabilities`
 - Android Option: `debug.vulkan.khronos_profiles.simulate_capabilities`
-- Options: SIMULATE_API_VERSION_BIT, SIMULATE_FEATURES_BIT, SIMULATE_PROPERTIES_BIT, SIMULATE_EXTENSIONS_BIT, SIMULATE_FORMATS_BIT, SIMULATE_FORMAT_PROPERTIES_BIT
-- Default Value: SIMULATE_API_VERSION_BIT, SIMULATE_FEATURES_BIT, SIMULATE_PROPERTIES_BIT
+- Options: SIMULATE_API_VERSION_BIT, SIMULATE_EXTENSIONS_BIT, SIMULATE_FEATURES_BIT, SIMULATE_PROPERTIES_BIT, SIMULATE_FORMATS_BIT
+- Default Value: Not set
```
christophe-lunarg, line 114: This is not correct but maybe that means we have a bug.

The manifest file specify the default values

christophe-lunarg (changes requested):

christophe-lunarg (approved):

layer/TUTORIAL.md
ziga-lunarg: My problem right now is how can I set `simulate_capabilities` to nothing? Do we need something like `SIMULATE_NONE_BIT`? Because if I set the env variable to empty, the default will be used.

christophe-lunarg: Ah, great point!

I think we should be able to set it to nothing... At least the validation layer has settings that works this way so it would be nice to be consistent on this.

PR #35, opened 2022-01-26
Various layer issues encountered during Android port

mark-lunarg:
Mostly minor stuff -- some tweaking of the json loader/parser arg and fixed some API names.

### DIFF
--- a/layer/profiles.cpp
+++ b/layer/profiles.cpp
@@ -4043,6 +4043,11 @@ VkResult JsonLoader::ReadProfile(const Json::Value root, const std::vector<std::
 static Json::Value ParseJsonFile(std::string filename) {
     Json::Value root = Json::nullValue;
 
+    // Remove newline from filename
+    if (int(filename.back() == 0xa) ) {
+        filename.pop_back();
+    }
+
     std::ifstream file;
     file.open(filename.c_str());
     if (!file.is_open()) {
--- a/layer/profiles.cpp
+++ b/layer/profiles.cpp
@@ -10378,9 +10378,9 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
                     feature_chain.pNext = &(pdd.physical_device_vulkan_1_3_features_);
                 }
 
-                dt->GetPhysicalDeviceProperties2KHR(physical_device, &property_chain);
-                dt->GetPhysicalDeviceFeatures2KHR(physical_device, &feature_chain);
-                dt->GetPhysicalDeviceMemoryProperties2KHR(physical_device, &memory_chain);
+                dt->GetPhysicalDeviceProperties2(physical_device, &property_chain);
+                dt->GetPhysicalDeviceFeatures2(physical_device, &feature_chain);
+                dt->GetPhysicalDeviceMemoryProperties2(physical_device, &memory_chain);
 
                 pdd.physical_device_properties_ = property_chain.properties;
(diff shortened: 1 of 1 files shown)


layer/profiles.cpp
```diff
@@ -10373,9 +10378,9 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
                     feature_chain.pNext = &(pdd.physical_device_vulkan_1_3_features_);
                 }
 
-                dt->GetPhysicalDeviceProperties2KHR(physical_device, &property_chain);
-                dt->GetPhysicalDeviceFeatures2KHR(physical_device, &feature_chain);
-                dt->GetPhysicalDeviceMemoryProperties2KHR(physical_device, &memory_chain);
+                dt->GetPhysicalDeviceProperties2(physical_device, &property_chain);
+                dt->GetPhysicalDeviceFeatures2(physical_device, &feature_chain);
+                dt->GetPhysicalDeviceMemoryProperties2(physical_device, &memory_chain);
```
christophe-lunarg, line 10383: Is it really what we should use? Why not the extension?
This implies Vulkan 1.1 requirements instead of 1.0 + get properties 2 just on this line. This said we didn't clarified the requirements for the Profiles layer so... unsure. 

@ziga-lunarg any idea on this?

mark-lunarg: The KHR APIs are part of an extension.  @mikes-lunarg would know for sure, but I believe it is invalid to use the KHR version of a promoted API unless the extension is also enabled.  Certainly, the Android Loader balks if you try to do so.

shoecart: For maximum compatibility you probably need to be able to call either one depending on how the application is initialized. If the enumerated instance version >= 1.1 then it is safe to call `GetPhysicalDeviceMemoryProperties2`, if the VK_KHR_get_physical_device_properties2 extension is enabled then it is safe to call `GetPhysicalDeviceMemoryProperties2KHR`. Right now the layer is using the `get_physical_device_properties2_active` variable to cover both cases (and using `VkApplicationInfo::apiVersion` instead of `vkEnumerateInstanceVersion`).

If I had to pick one version for expediency, I would choose `GetPhysicalDeviceProperties2` because hardly anyone is running around with a 1.0 instance (ie loader).

mikes-lunarg: For maximum compatibility you probably need to be able to call either one depending on how the application is initialized. If the enumerated instance version >= 1.1 then it is safe to call `GetPhysicalDeviceMemoryProperties2`, if the VK_KHR_get_physical_device_properties2 extension is enabled then it is safe to call `GetPhysicalDeviceMemoryProperties2KHR`. Right now the layer is using the `get_physical_device_properties2_active` variable to cover both cases (and using `VkApplicationInfo::apiVersion` instead of `vkEnumerateInstanceVersion`).

If I had to pick one version for expediency, I would choose `GetPhysicalDeviceProperties2` because hardly anyone is running around with a 1.0 instance (ie loader).

ziga-lunarg: But we have a requirement to only use Vulkan 1.0 + `VK_KHR_get_physical_device_properties2`, so I need we will have to enable the extension.

christophe-lunarg: `Vulkan 1.0 + VK_KHR_get_physical_device_properties2` this is for the Profiles API library but maybe we can require Vulkan 1.1 in the layer. This said, I think we just need to be clear about the requirement and document it in the layer doc.

PR #38, opened 2022-01-26
profiles layer: Fix getting SDK env variable

ziga-lunarg:
Fix checking if env variable is not set.


richard-lunarg:
This did fix the SEGFAULT on the four failing tests!

PR #46, opened 2022-01-27
layer: fix API version check

christophe-lunarg:


layer/profiles.cpp
```diff
@@ -1089,12 +1089,6 @@ class PhysicalDeviceData {
 
     static void Destroy(const VkPhysicalDevice pd) {
         LogMessage(DEBUG_REPORT_DEBUG_BIT, "PhysicalDeviceData::Destroy()\n");
-
-        auto pdd = Find(pd);
-        assert(pdd);
```
ziga-lunarg, line 1094: Maybe this assert can remain, it was in here before this was added.

richard-lunarg:
All tests pass on Apple Silicon!

PR #61, opened 2022-01-31
Create Android test package for `VpLibrary_test_api_create_device_android`

mark-lunarg:
Added infrastructure and associated files to build the `VpLibrary_test_api_create_device_android` test for Android along with an installable test package.  The resulting APK can then be installed and run on an Android device.

Initial cmake command (see BUILD.md for specifics):

```
cmake -Bandroid-build \
  -DCMAKE_TOOLCHAIN_FILE=/home/mark/gdev/android-ndk-r21d/build/cmake/android.toolchain.cmake \
  -DANDROID_PLATFORM=28 \
  -DANDROID_BUILD_TOOLS=30.0.3 \
  -DANDROID_ABI=arm64-v8a
```

Change into the android-build dir and run:

`cmake --build . --parallel`

Install the APK as follows:

`adb install -r apk/out/VpLibrary_test_api_create_device_android.apk`

And run the test:

`adb shell am start -a android.intent.MAIN -c android-intent.category.LAUNCH -n com.example.VpLibrary_test_api_create_device_android/android.app.NativeActivity --es args --gtest_catch_exceptions=0`

Test output goes to logcat, which you can pull and scrape looking for `VpLibrary_test_api_create_device_android:`.  You should observe something like the following:

```
01-31 14:01:51.710  I VpLibrary_test_api_create_device_android: Intent args = --gtest_catch_exceptions=0
01-31 14:01:51.710  D VpLibrary_test_api_create_device_android: filter = --gtest_catch_exceptions=0
01-31 14:01:51.724  I VpLibrary_test_api_create_device_android: [ RUN      ] library_api.vpCreateDevice
01-31 14:01:51.735  I VpLibrary_test_api_create_device_android: [       OK ]library_api.vpCreateDevice (10 ms)
01-31 14:01:51.735  I VpLibrary_test_api_create_device_android: [ RUN      ] library_api.vpGetPhysicalDeviceProfileSupport
01-31 14:01:51.738  I VpLibrary_test_api_create_device_android: [       OK ]library_api.vpGetPhysicalDeviceProfileSupport (3 ms)
01-31 14:01:51.738  I VpLibrary_test_api_create_device_android: ==== Tests PASSED ====
```


lunarpapillo:
I'm working on corresponding CI in *VulkanTests* to execute this package.

christophe-lunarg (approved):

ziga-lunarg (approved):
I don't understand much of it, but looks good.

library/test/test_api_create_device_android.cpp
```diff
@@ -24,6 +25,233 @@
 #include <vulkan/vulkan_android.h>
 #include <vulkan/debug/vulkan_profiles.hpp>
 
+#if defined(ANDROID)
```
lunarpapillo, line 28: Similar/identical boilerplate will have to be added to every test `.cpp` file,  yes?  Both here and in other projects that try to use the CMake Android support...

I'm wondering whether there's a way to package this in such a way as to make it easier to use for many tests and in many projects... but it's late and I'm stupid and I'm not coming up with anything workable...

library/test/platforms/android/res/values/strings.xml
```diff
@@ -0,0 +1,24 @@
...
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+
+<!-- This file contains resource definitions for displayed strings, allowing
+     them to be changed based on the locale and options. -->
+
+<resources>
+    <!-- Simple strings. -->
+    <string name="app_name">VpLibrary_test_api_create_device_android</string>
```
lunarpapillo, line 22: This file, by its comments, seems intended to allow strings to change by locale.  We don't really care about that... is it feasible at all to remove this file and just use a constant string somewhere, or is something like this required for all Android builds?

lunarpapillo (approved):
What strange and wonderful magic...

lunarpapillo:
In order to pass CI, this change will need the **VulkanTests** change to ensure a valid `ANDROID_BUILD_TOOLS` is passed to the build:  https://github.com/LunarG/VulkanTests/pull/206

Once that change is merged, we can re-run CI on this change and get it merged.

Note that there is an issue, that CI on Android will currently fail if actually run: https://github.com/KhronosGroup/Vulkan-Profiles/issues/87

But this change can be safely merged before that issue is addressed, as long as we maintain the `--no-test` switch on Android CI runs (as we currently do).

Once that issue is resolved, we should be able to remove the `--no-test` switch and actually get Android build and execution on CI runs.

library/test/platforms/android/res/values/strings.xml
mark-lunarg: Just did some research and this was tied to the 'label' attribute, which is needed for a short text string associated with an icon on the device.  Since we are only running the APKs through ADB I was able to remove these files and references to them and things seem to be working.  Sweet.

PR #72, opened 2022-02-03
Call For Feedback:  Add profiles layer settings struct for Layer Tests

mark-lunarg:
Reworked settings on layer tests to use a programmatic method instead of relying on environment variables.

This change allows tests to add a structure containing environment variable settings to the `CreateInstance()` pNext chain.

Incorporated review feedback, other than the layer path, which will need to be addressed with and through CI.

<answered, mostly>
Open questions:

- Include all settings or just the relevant layer-specific ones?
- Struct names, default constructor, namespace issues?
- Default values for settings in `VkProfileLayerSettingsEXT` structure
- Left VK_LAYER_PATH in tests -- it is unusual to set this variable _inside_ the tests themselves.  Typically it is set before (and outside) running the tests to control which layer is getting tested.  It is also a no-op on Android.
- Significant overlap of the `LayerSettings` and the `VkProfileLayerSettingsEXT` structures

Basically, is this suitable for using across all the layer tests or are modifications or improvements desired?  All feedback is good feedback.


ziga-lunarg (approved):

ziga-lunarg:
Default values should be what the manifest says they are. 
I agree that `VK_LAYER_PATH` would make more sense to be set in cmake. 
I also think it would be sense for `VkProfileLayerSettingsEXT` to have everything `LayerSettings` has, since we would want to test all the options.

mark-lunarg:
@ziga-lunarg, thanks, this is exactly the kind of stuff I was looking for.  For the default values, should I be looking at the specified manifest each time?

ziga-lunarg:
@mark-lunarg Yes, or you can just look into layer/TUTORIAL.md, default values are also documented there.

mark-lunarg:
@ziga-lunarg, I am setting the defaults in the settings constructor, but I see some of the tests set the default values explicity (i.e., emulate_portability = true;).  Can I (or should I) scrub these from the tests?  Seems like it'd be a good idea but I don't want to step on any toes.

ziga-lunarg:
@mark-lunarg Yes, that should not change the settings, if it's already the default, so it would make sense to remove it.

ziga-lunarg:
Much better!

PR #74, opened 2022-02-04
profiles layer: Print all warnings before failing

ziga-lunarg:
When `fail_on_error` setting is used all warnings should be printed before failing.


mark-lunarg (approved):
Sheesh, this is so much better.  Nice.

PR #75, opened 2022-02-04
profiles layer: Implement GetQueueFamilyProperties

ziga-lunarg:


christophe-lunarg (approved):

layer/profiles.cpp
```diff
@@ -3688,6 +3917,153 @@ bool JsonLoader::GetFormat(const Json::Value &formats, const std::string &format
...
 }
 
+bool QueueFamilyMatch(const VkQueueFamilyProperties &device_qfp, const VkQueueFamilyProperties &profile_qfp) {
+    if ((device_qfp.queueFlags & profile_qfp.queueFlags) != profile_qfp.queueFlags) {
+        return false;
+    }
+    if (device_qfp.queueCount < profile_qfp.queueCount) {
+        return false;
+    }
+    if (device_qfp.timestampValidBits != profile_qfp.timestampValidBits) {
```
aqnuep, line 3927: This should be `<`, not `!=`. While the `vk.xml` is missing the corresponding limit type this is actually a minimum maximum.

layer/profiles.cpp
```diff
@@ -3688,6 +3917,153 @@ bool JsonLoader::GetFormat(const Json::Value &formats, const std::string &format
...
+        return false;
+    }
+    if (device_qfp.minImageTransferGranularity.width >= profile_qfp.minImageTransferGranularity.width ||
+        device_qfp.minImageTransferGranularity.height >= profile_qfp.minImageTransferGranularity.height ||
+        device_qfp.minImageTransferGranularity.depth >= profile_qfp.minImageTransferGranularity.depth) {
```
aqnuep, line 3932: These should be `>`, not `>=`.

layer/profiles.cpp
```diff
@@ -3688,6 +3917,153 @@ bool JsonLoader::GetFormat(const Json::Value &formats, const std::string &format
...
+    }
+    return true;
+}
+
+bool JsonLoader::GetQueueFamilyProperties(const Json::Value &qf_props, ArrayOfVkQueueFamilyProperties *dest,
+                                          VkQueueFamilyGlobalPriorityPropertiesKHR *gp_dest,
+                                          VkVideoQueueFamilyProperties2KHR *vqfp_dest,
+                                          VkQueueFamilyCheckpointPropertiesNV *qfcp_dest,
+                                          VkQueueFamilyCheckpointProperties2NV *qfcp2_dest,
+                                          VkQueueFamilyQueryResultStatusProperties2KHR *qfqrsp_dest) {
```
aqnuep, line 3943: Maybe I'm missing something, but why the `VkQueueFamilyProperties` parameter is an array while the others aren't?
Each should be, as each queue family can have different global priorities, checkpoint properties, etc.

layer/profiles.cpp
```diff
@@ -3688,6 +3917,153 @@ bool JsonLoader::GetFormat(const Json::Value &formats, const std::string &format
...
+            qfp.sType = VK_STRUCTURE_TYPE_QUEUE_FAMILY_PROPERTIES_2;
+            for (const auto &feature : props["queueFlags"]) {
+                qfp.queueFamilyProperties.queueFlags |= StringToVkQueueFlags(feature.asString());
+            }
+            qfp.queueFamilyProperties.queueCount = props["queueCount"].asInt();
+            qfp.queueFamilyProperties.timestampValidBits = props["timestampValidBits"].asUInt();
+            const auto &minImagetransferGranularity = props["minImageTransferGranularity"];
+            qfp.queueFamilyProperties.minImageTransferGranularity.width = minImagetransferGranularity["width"].asUInt();
+            qfp.queueFamilyProperties.minImageTransferGranularity.height = minImagetransferGranularity["height"].asUInt();
+            qfp.queueFamilyProperties.minImageTransferGranularity.depth = minImagetransferGranularity["depth"].asUInt();
```
aqnuep, line 3957: What happens when the profile doesn't define a particular field of the structure?
Not sure what's the intention in the end. Should we just return zeros for those? Does the above code actually do that if the corresponding member isn't actually present in the JSON data, or will to raise an exception?
This actually leads to the wider question of what the default values of members in structs should be when the profile doesn't define a value. If zeros make sense always, then that's okay, but I'm not quite sure.

layer/profiles.cpp
```diff
@@ -8623,6 +9014,116 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumerateDeviceExtensionProperties(VkPhysicalDevi
...
+
+    // Careful: cannot use EnumerateProperties() here! (because src and dst structs are not the same type)
+    const uint32_t copy_count = (*pQueueFamilyPropertyCount < src_count) ? *pQueueFamilyPropertyCount : src_count;
+    const VkQueueFamilyProperties2 *src_props = pdd->arrayof_queue_family_properties_.data();
+    for (uint32_t i = 0; i < copy_count; ++i) {
+        pQueueFamilyProperties2[i].queueFamilyProperties = src_props[i].queueFamilyProperties;
+    }
+    *pQueueFamilyPropertyCount = copy_count;
+    for (uint32_t i = 0; i < copy_count; ++i) {
+        FillQueueFamilyPropertiesPNextChain(pdd, pQueueFamilyProperties2[i].pNext);
```
aqnuep, line 9117: Indeed, it does look like you're filling the same extension properties for all queue families, even though technically each could have their own set of extension property structure data, different from the ones for other queue families.
You need to make all the structure data containers arrays so each queue family could have their own set of values.

profiles/test/data/VP_LUNARG_test_api.json
```diff
@@ -3000,6 +3013,14 @@
...
                             "VK_PIPELINE_STAGE_2_DRAW_INDIRECT_BIT"
                         ]
+                    },
+                    "VkQueueFamilyCheckpointPropertiesNV": {
+                        "checkpointExecutionStageMask": [
+                            "VK_PIPELINE_STAGE_TRANSFER_BIT"
+                        ]
+                    },
+                    "VkQueueFamilyQueryResultStatusProperties2KHR": {
+                        "supported": true
```
aqnuep, line 3023: You'll need to add some test JSON with multiple queue families, each with its own set of extension properties to make sure that once you've changed the code to handle separate set of extension properties then they each get their own sets of values.

aqnuep (changes requested):

layer/profiles.cpp
```diff
@@ -3688,6 +3917,153 @@ bool JsonLoader::GetFormat(const Json::Value &formats, const std::string &format
...
+            *qfqrsp_dest = qfqrsp;
+        }
+    }
+
+    bool valid = true;
+
+    for (const auto& profile_qfp : *dest) {
+        bool supported = false;
+        for (const auto& device_qfp : pdd_.device_queue_family_properties_) {
+            if (QueueFamilyMatch(device_qfp.queueFamilyProperties, profile_qfp.queueFamilyProperties)) {
```
aqnuep, line 4018: I don't see you checking anywhere that the extension properties match, you only check the base structure, while in fact the extension properties may define additional requirements, not just data that should be reported back.

layer/profiles.cpp
ziga-lunarg: I was told there can only be one of each in the pNext chain, duplicates are considered an error.

layer/profiles.cpp
ziga-lunarg: Perhaps the values should then be taken from the compatible queue that will be used?

layer/profiles.cpp
ziga-lunarg: The way I see it is that the extensions are separate from the actual queue family properties, and I check the values below this code.

layer/profiles.cpp
aqnuep: Well, that's not what the spec says, also, the schema allows defining separate structures for each queue family definition index (it always did, even before it was generated).

What you probably refer to is that a single `pNext` chain can contain a structure only once. However, we're talking about multiple `pNext` chains here, one per queue family, each can have its own version of any of the chained structures.

layer/profiles.cpp
aqnuep: That's what I'm not sure about, as in that case the layer would actually expose more capabilities than what's defined in the profile. So probably zero is fine. Still, do these actually work when the property is missing from the JSON? I'm not familiar with the behavior of the JSON library you use, so I wonder whether trying to reference e.g. `["width"]` will simply return default value (e.g. zero) or will raise an exception or something.

layer/profiles.cpp
aqnuep: No, this is the same problem as the one you've misunderstood above. Each base structure can have its own extension chain, so you actually have to check each chain in each iteration separately.

profiles/test/data/VP_LUNARG_test_api.json
```diff
@@ -2983,13 +2983,26 @@
...
+                    "VkQueueFamilyProperties2": {
+                        "queueFamilyProperties": {
+                            "queueFlags": [ "VK_QUEUE_COMPUTE_BIT", "VK_QUEUE_TRANSFER_BIT" ],
+                            "queueCount": 382,
+                            "timestampValidBits": 383,
+                            "minImageTransferGranularity": {
+                                "width": 384,
+                                "height": 385,
+                                "depth": 386
+                            }
```
aqnuep, line 2995: You should also add some tests in which not all fields are defined, e.g. `VkQueueFamilyProperties` without `timestampValidBits`, or something, just to make sure that parsing works, even when partial structure data are provided.

I guess in such cases the `timestampValidBits` should actually come from the real device's queue family properties, and just the properties defined by the profile should be overridden, but I'm not sure this was thought through before, so maybe we can just default it to zero.

Nonetheless, the layer should be able to handle in some way when some fields aren't defined by the profile.

aqnuep (commented):
This looks much better now, but I believe there's still something wrong with it (see my first comment above).

layer/profiles.cpp
```diff
@@ -4056,6 +4290,248 @@ bool JsonLoader::valid(ExtensionSupport support) {
...
+                if (permutations[i] < qfp->size()) {
+                    ordered.push_back((*qfp)[permutations[i]]);
+                } else {
+                    ordered.push_back(QueueFamilyProperties());
+                }
+            }
+            *qfp = ordered;
+            return true;
+        }
+    } while (std::next_permutation(permutations.begin(), permutations.end()));
```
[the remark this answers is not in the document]
ziga-lunarg: This checks that all requested queues can actually be used and allows the layer to reorder them so that the profile queue will be on the same index that the device queue that supports it is.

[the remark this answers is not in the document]
aqnuep: Yes, I got that. I just don't think you have to test for every permutation, which requires `O(n * n!)` time complexity, as it should be doable in `O(n^2)`, I believe.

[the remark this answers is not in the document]
aqnuep: Though probably it doesn't matter as `n` won't ever be really large.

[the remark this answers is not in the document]
ziga-lunarg: I thought about time complexity too, but this code will be hard to test, so I went with the safe version for now.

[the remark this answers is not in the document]
aqnuep: Yes, I concur. I've tweaked the library code to work similarly (see PR #129). If you could review that, I'd appreciate it.

aqnuep (approved):

PR #80, opened 2022-02-06
profiles layer: Add GetPhysDev2 extensions of not present

ziga-lunarg:
The Profiles Layer requires the `VK_KHR_get_physical_device_properties2` extension, without it it's not functional.


christophe-lunarg (approved):
Great, I think it needs doc update too to clarify the requirements publicly.

PR #82, opened 2022-02-07
profiles layer: Use KHR version of functions when vulkan version is 1.0

ziga-lunarg:


mark-lunarg (approved):
Oops!  That was probably me.

PR #97, opened 2022-02-09
profiles layer: Transfer device values from extension structs to promoted structs

ziga-lunarg:
If instance version is not high enough structs such as `VkPhysicalDeviceVulkan11Features` or `VkPhysicalDeviceVulkan12Properties` might not be loaded, but the extensions always will if the extension is present on the device. So it makes sense to copy them from those structs. Not sure what to do about the members which are not part of any extension, will be addressed in a separate PR in the future.


christophe-lunarg (commented):
Probably good enough for this release

PR #98, opened 2022-02-09
profiles layer: Improve message logs for profile and device values

ziga-lunarg:
JSON value => Profile value 
and
existing value => Device value


christophe-lunarg (approved):

christophe-lunarg:
Nice!

PR #103, opened 2022-02-09
profiles layer: Skip loading structs from excluded extensions

### LINKED ISSUE #96: Layer: Excluded extensions are not reported?
image

i would expect at least a warning or notification that say that we are excluding an extension that is required by the selected profile.

ziga-lunarg:
Closes #96

### DIFF
--- a/layer/profiles.cpp
+++ b/layer/profiles.cpp
@@ -1926,10 +1926,17 @@ class JsonLoader {
 
    private:
     std::uint32_t profile_api_version;
+    std::vector<std::string> excluded_extensions;
 
     struct Extension {
         std::string name;
         int specVersion;
+    };
+
+    enum ExtensionSupport {
+        UNSUPPORTED,
+        EXCLUDED,
+        SUPPORTED,
     };
 
     bool WarnDuplicatedFeature(const Json::Value &parent);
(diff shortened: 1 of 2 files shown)


aqnuep:
Some of these aren't just extension structures, but ones that also got promoted to core versions, e.g. `VkPhysicalDeviceMaintenance3Properties`.

Is it intended to skip these structures if the core API version would include them even without the excluded extensions?

christophe-lunarg:
it's not clear to me what this PR is doing. i build the branch and tried it:
profiles_excluded_ext

```
PROFILES WARNING: shaderDrawParameters JSON value is enabled in the profile, but the device does not support it.
```
Is still appearing...
And we don't get any notification regarding the features of `VK_KHR_8bit_storage` that is excluded and in the profile so it's not clear for the Vulkan developer to understand what's going on. 

I think the exclude settings should override the profile but this should be notified to the Vulkan developer (not only in the settings notification, but in how the settings are effectively executed).

ziga-lunarg:
@aqnuep I'm not sure what I should do there, it probably shouldn't be excluded

ziga-lunarg:
@christophe-lunarg Added the logs now

PR #107, opened 2022-02-10
profiles layer: Add NONE simulate capability flag

ziga-lunarg:
This is needed for tests.

### DIFF
--- a/layer/profiles.h
+++ b/layer/profiles.h
@@ -21,12 +21,12 @@
 #include "vk_layer_settings.h"
 
 enum SimulateCapabilityFlag {
+    SIMULATE_NONE_BIT = 0,
     SIMULATE_API_VERSION_BIT = 1 << 0,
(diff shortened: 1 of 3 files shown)


ziga-lunarg:
> `SIMULATE_API_VERSION_BIT` is currently `1`

You're right, I must have been debugging wrong, but I still need it for tests.

mark-lunarg:
@ziga-lunarg, if you use SIMULATE_ALL_CAPABILITIES, then it will set the NONE bit with all the others, yes?  Does setting it to zero not work?

layer/profiles.h
```diff
@@ -21,12 +21,12 @@
 #include "vk_layer_settings.h"
 
 enum SimulateCapabilityFlag {
+    SIMULATE_NONE_BIT = 0,
```
christophe-lunarg, line 24: It's not a `SIMULATE_NONE_BIT ` but `SIMULATE_NONE` if that's necessary.

PR #115, opened 2022-02-11
profiles layer: Fix checking features and properties support

ziga-lunarg:
Promoted struct should be checked against the vulkan API version instead of extension. 
Effective API version is determined as in validation layers. 
Removed TransferDeviceValues as it's no longer needed, since all checks are correct and proper warnings will be emitted.
#103 will have to be rebased and updated from this


aqnuep (commented):
This looks reasonable to me.

PR #132, opened 2022-02-14
profiles layer: Add script to generate a test profile using all extensions

ziga-lunarg:


layer/scripts/gen_all_ext_profile.py
```diff
@@ -0,0 +1,530 @@
+import os
+import xml.etree.ElementTree as etree
+import math
+import argparse
+from fractions import Fraction
+
+PROFILES = '''
+    "profiles": {
```
aqnuep, line 8: My general comment would be that, instead of using strings, you should prefer constructing JSONs in python by simply using python dicts and lists and then dump the formatted JSON output using `json.dumps(data, indent=4)`.

You can check it out how that works in the `VulkanProfilesSchemaGenerator` of `genvp.py`. In particular, you can check the `gen_schema` method which contains the dict/list based skeleton of the generated JSON schema.

layer/scripts/gen_all_ext_profile.py
```diff
@@ -0,0 +1,530 @@
...
+        gen += '}\n'
+        return gen
+
+class VulkanExtension():
+    def __init__(self, spec_version, feature, property):
+        self.spec_version = spec_version
+        self.feature = feature
+        self.property = property
+
+class VulkanRegistry():
```
aqnuep, line 380: It seems like you reused part of the infrastructure in `genvp.py`. I think maybe a better idea would be to factor out the reusable components (e.g. `VulkanRegistry` and related classes), unless you intentionally departed from some of the data layouts used there.

christophe-lunarg (changes requested):
Where is Where is gen_all_ext_profile.py executed? I am not seeing any CMake changes to do this but I might be missing something.

christophe-lunarg:
It would be great to use default values (minimum requirements) for the capabilities read from vk.xml but these values are not present in vk.xml...

profiles/test/data/VP_LUNARG_test_api_generated.json
```diff
@@ -0,0 +1,3055 @@
...
+                    "minSgprAllocation": 42,
+                    "maxSgprAllocation": 43,
+                    "sgprAllocationGranularity": 44,
+                    "vgprsPerSimd": 45,
+                    "minVgprAllocation": 46,
+                    "maxVgprAllocation": 47,
+                    "vgprAllocationGranularity": 48
+                },
+                "VkPhysicalDeviceShaderCoreProperties2AMD": {
+                    "shaderCoreFeatures": [],
```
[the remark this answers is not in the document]
ziga-lunarg: There are no valid flags for `VkShaderCorePropertiesFlagsAMD` at the moment, the only enum is `VK_SHADER_CORE_PROPERTIES_FLAG_BITS_MAX_ENUM_AMD = 0x7FFFFFFF`

christophe-lunarg (approved):

PR #133, opened 2022-02-14
profiles layer: Some updates for tutorial

ziga-lunarg:


layer/TUTORIAL.md
```diff
@@ -196,6 +200,7 @@ Prevents device extensions from being reported by the Vulkan physical device.
 
 Prevents image formats from being reported as supported by the Vulkan physical device.
 
+**Note:** If multiple values are used they are separated by ','. On windows they can be separated by ';' as well and with ':' on other platforms.
```
christophe-lunarg, line 203: If multiple values are used they are separated by ',' is actually for the layer_settings.txt file

While On windows they can be separated by ';' as well and with ':' on other platforms. is for env variables usage.

christophe-lunarg (approved):

lunarpapillo (approved):
I echo Christophe's observation, but otherwise this looks great! :-)

PR #137, opened 2022-02-15
Auto-generate schema version from header version

### LINKED ISSUE #135: schema: The $id and title are not generated using the Vulkan Header API version
The generated schema file has the following content:
```
    "$id": "https://schema.khronos.org/vulkan/profiles-1.3.203.json#",
    "title": "Vulkan Profiles Schema for Vulkan 1.3.203",
```
But it's not correct because the current Vulkan Header is 1.3.204, hence we should have:
```
```
    "$id": "https://schema.khronos.org/vulkan/profiles-1.3.204.json#",
    "title": "Vulkan Profiles Schema for Vulkan 1.3.204",
```
So it would be great to generate these lines using the Vulkan Header API version instead of these being hard coded.

aqnuep:
Closes #135.

### DIFF
--- a/schema/profile_schema.json
+++ b/schema/profile_schema.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://schema.khronos.org/vulkan/profiles-1.3.203.json#",
-    "title": "Vulkan Profiles Schema for Vulkan 1.3.203",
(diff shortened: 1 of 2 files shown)


christophe-lunarg (approved):

christophe-lunarg:
Oups I forget to wait for C.I. to finish...

christophe-lunarg:
C.I. Passed \o/

PR #139, opened 2022-02-15
Fix a handful of compiler warnings in the layer and set the  Werror flag for non-Windows compilers

mark-lunarg:
There was a growing list of warning spew on Linux and Android builds.  Fixed these and turned on the `Werror` flag.


ziga-lunarg (approved):
LGTM

PR #140, opened 2022-02-16
profiles layer: Set instance version from profile

### LINKED ISSUE #118: Allow overriding the instance application apiVersion when SIMULATE_API_VERSION_BIT is set
Currently, this can't be done because the Profiles file is not loaded when calling CreateInstance but this should be resolve applying Profiles with higher Vulkan version requirements than the Vulkan application.

This is an item for SDK 1.3.204.1

ziga-lunarg:
Closes #118

### DIFF
--- a/layer/profiles.cpp
+++ b/layer/profiles.cpp
@@ -4970,35 +4970,38 @@ struct JsonValidator {
 };
 
 VkResult JsonLoader::LoadFile(std::string filename) {
+    LogMessage(DEBUG_REPORT_DEBUG_BIT, format("JsonLoader::LoadFile(\"%s\")\n", filename.c_str()));
+
     profile_filename_ = filename;
     if (filename.empty()) {
         if (!layer_settings->profile_name.empty()) {
             LogMessage(DEBUG_REPORT_WARNING_BIT,
-                       format("Profile name is set to \"%s\", but profile_file is unset. The profile will not be loaded.\n",
-                              layer_settings->profile_name.c_str()));
+                format("Profile name is set to \"%s\", but profile_file is unset. The profile will not be loaded.\n",
+                layer_settings->profile_name.c_str()));
         }
         return VK_SUCCESS;
     }
     std::ifstream json_file(filename);
     if (!json_file) {
-        LogMessage(DEBUG_REPORT_ERROR_BIT, format("JsonLoader failed to open file \"%s\"\n", filename.c_str()));
+        LogMessage(DEBUG_REPORT_ERROR_BIT, 
+            format("Fail to open file \"%s\"\n", filename.c_str()));
         return layer_settings->debug_fail_on_error ? VK_ERROR_INITIALIZATION_FAILED : VK_SUCCESS;
     }
-
-    LogMessage(DEBUG_REPORT_DEBUG_BIT, format("JsonLoader::LoadFile(\"%s\")\n", filename.c_str()));
 
     Json::Value root = Json::nullValue;
     Json::CharReaderBuilder builder;
     std::string errs;
     bool success = Json::parseFromStream(builder, json_file, &root_, &errs);
     if (!success) {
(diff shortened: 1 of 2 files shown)


layer/profiles.cpp
```diff
@@ -7553,55 +7607,95 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
...
+    bool changed_version = false;
+    uint32_t profile_api_version = json_loader.GetProfileApiVersion();
+    if (VK_VERSION_MAJOR(requested_version) < VK_VERSION_MAJOR(profile_api_version) ||
+        VK_VERSION_MINOR(requested_version) < VK_VERSION_MINOR(profile_api_version)) {
+        LogMessage(DEBUG_REPORT_WARNING_BIT,
+                   ::format("The Profiles Layer requires API version %s, but instance is being created "
+                            "with API version %s. The API version will be set to %s.\n",
+                            StringAPIVersion(requested_version).c_str(), StringAPIVersion(profile_api_version).c_str(),
+                            StringAPIVersion(requested_version).c_str()));
+        requested_version = profile_api_version;
```
ziga-lunarg: Should this be under a setting? `SIMULATE_API_VERSION_BIT` or a new one?

layer/profiles.cpp
```diff
@@ -4680,40 +4715,40 @@ VkResult JsonLoader::ReadProfile(const Json::Value root, const std::vector<std::
                 AddPromotedExtensions(properties_api_version);
             }
         } else if (layer_settings->simulate_capabilities & SIMULATE_API_VERSION_BIT) {
-            AddPromotedExtensions(this->profile_api_version);
+            AddPromotedExtensions(this->profile_api_version_);
         }
 
-        if (VK_VERSION_PATCH(this->profile_api_version) > VK_VERSION_PATCH(pdd_.physical_device_properties_.apiVersion)) {
+        if (VK_VERSION_PATCH(this->profile_api_version_) > VK_VERSION_PATCH(pdd_->physical_device_properties_.apiVersion)) {
```
aqnuep: `VK_VERSION_PATCH` (actually deprecated, and `VK_API_VERSION_PATCH` should be used instead) is only returning the patch version from the full version. Are you sure that's what you meant to use here and not compare the major+minor version instead?

layer/profiles.cpp
```diff
@@ -4954,39 +4990,66 @@ VkResult JsonLoader::LoadFile(std::string filename) {
...
+
+    /*if (VK_VERSION_MINOR(this->profile_api_version_) > VK_VERSION_MINOR(requested_version)) {
+        LogMessage(DEBUG_REPORT_WARNING_BIT,
+                   format("The application requested a Vulkan 1.%d instance however the selected profile requires Vulkan 1.%d. "
+                          "Some profile capabilities might not be applied, we recommend changing the Vulkan application to request "
+                          "a Vulkan 1.%d instance.\n",
+                          VK_VERSION_MINOR(requested_version), VK_VERSION_MINOR(this->profile_api_version_),
+                          VK_VERSION_MINOR(this->profile_api_version_))
+                       .c_str());
+    }*/
```
aqnuep: I guess this commented out code should be deleted.

aqnuep (commented):
I can't tell whether this behaves as expected, as the code interactions are non-trivial, so it would be great to have some sort of test for this.

As discussed, we'd probably need some sort of Vulkan API mock to be able to catch what gets passed to `vkCreateInstance`, although it's not trivial how to do it in a setting like this, where we'd actually have to capture `vkCreateInstance` within the layer chain.

Thus, personally, I'd be happy enough if this is verified manually for now if an automated test seems unfeasible.

christophe-lunarg (approved):
I made some fixes:
- Improved messages
- Protected application API version override with SIMULATE_API_VERSION_BIT
- Fixed usage of deprecated Vulkan macros
- When a profile name is not found, it's just a warning and we load the default profile

PR #147, opened 2022-02-21
generator script: Update finding enums

ziga-lunarg:
This makes sure only enums from core and vulkan extensions are found

### DIFF
--- a/scripts/genvp.py
+++ b/scripts/genvp.py
@@ -1382,7 +1382,7 @@ class VulkanRegistry():
 
     def parseVersionInfo(self, xml):
         self.versions = dict()
-        for feature in xml.findall('./feature'):
+        for feature in xml.findall("./feature[@api='vulkan']"):
             if re.search(r"^[1-9][0-9]*\.[0-9]+$", feature.get('number')):
                 self.versions[feature.get('name')] = VulkanVersion(feature)
             else:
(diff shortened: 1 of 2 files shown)


mark-lunarg (approved):

aqnuep:
The proposed changes seem fine, but you should also change all other places where the "feature" and "extension" elements are referenced in the XPaths.

aqnuep (changes requested):
Also, as these likely change e.g. the generated schema, it would be good to include any such generated content changes in the PR before merging.

scripts/genvp.py
```diff
@@ -1391,7 +1391,7 @@ def parseVersionInfo(self, xml):
 
     def parseExtensionInfo(self, xml):
         self.extensions = dict()
-        for ext in xml.findall("./extensions/extension"):
+        for ext in xml.findall("./extensions/extension[@supported='vulkan']"):
             # Only care about enabled extensions
             if ext.get('supported') == "vulkan":
```
aqnuep, line 1396: Now this explicit if-check on the attribute value can be eliminated as the XPath filters out any other possibilities.

aqnuep (commented):
Other than that minor refactoring comment, this looks fine.

scripts/genvp.py
ziga-lunarg: Good catch, thanks!

PR #148, opened 2022-02-21
FEEDBACK REQUESTED:  Introduce layer test framework for for layer tests

mark-lunarg:
Please look at changes made to **tests_capabilities.cpp** and **tests_promoted.cpp**.

Gtest will let you define classes for the tests that greatly simplify setup and teardown, and these classes can be shared among different test modules or used to build upon each other.  They also do complete setup/teardowns for each test, preventing any state from carrying through or between different tests.  This is necessary because the android changes require a single main() function and it must be used for all tests.

This example PR has some android stuff in it but that isn't implemented yet -- I did want to see if there is agreement that the changes made to the above test files seem reasonable -- the other files can be ignored for now.

### DIFF
--- a/layer/profiles.cpp
+++ b/layer/profiles.cpp
@@ -468,7 +468,8 @@ static std::string GetFormatFeature2String(VkFormatFeatureFlagBits2 flags) {
     return result;
 }
 
-static VkProfileLayerSettingsEXT *layer_settings;
+static VkProfileLayerSettingsEXT *layer_settings = new VkProfileLayerSettingsEXT{};
+
 static std::string GetQueueFlagsToString(VkQueueFlags flags) {
     std::string result = {};
 
@@ -7546,6 +7547,9 @@ static VkResult LayerSetupCreateInstance(const VkInstanceCreateInfo *pCreateInfo
 VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCallbacks *pAllocator,
                                               VkInstance *pInstance) {
     // This needs to be created before LogMessage is called, as it is dependent on these settings
+    if (layer_settings) {
+        delete layer_settings;
+    }
     layer_settings = new VkProfileLayerSettingsEXT;
 
     LogMessage(DEBUG_REPORT_DEBUG_BIT, "CreateInstance\n");
--- a/layer/tests/tests_capabilities.cpp
+++ b/layer/tests/tests_capabilities.cpp
@@ -33,17 +33,10 @@ static const char* CONFIG_PATH = "bin/Release";
 static const char* CONFIG_PATH = "lib";
 #endif
 
-struct TestInit {
-    TestInit() : instance(VK_NULL_HANDLE), physical_device(VK_NULL_HANDLE) {}
 
-    ~TestInit() {
-        if (this->instance != VK_NULL_HANDLE) {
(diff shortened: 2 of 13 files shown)


layer/tests/tests_promoted.cpp
```diff
@@ -33,17 +33,9 @@ static const char* CONFIG_PATH = "bin/Release";
...
-
-    ~TestInit() {
-        if (this->instance != VK_NULL_HANDLE) {
-            vkDestroyInstance(this->instance, nullptr);
-            this->instance = VK_NULL_HANDLE;
-        }
-    }
-
-    void init() {
+class layer_promoted : public VkTestFramework {
```
ziga-lunarg, line 36: Should it be `LayerPromoted`?

layer/tests/tests_capabilities.cpp
```diff
@@ -33,17 +33,10 @@ static const char* CONFIG_PATH = "bin/Release";
...
 
-    ~TestInit() {
-        if (this->instance != VK_NULL_HANDLE) {
-            vkDestroyInstance(this->instance, nullptr);
-            this->instance = VK_NULL_HANDLE;
-        }
-    }
-
-    void init() {
+class profiles : public VkTestFramework {
```
ziga-lunarg: Should profiles be capitalized?

ziga-lunarg (commented):
Looks good, but I am unfamiliar with android functions.

layer/profiles.cpp
```diff
@@ -7546,6 +7547,9 @@ static VkResult LayerSetupCreateInstance(const VkInstanceCreateInfo *pCreateInfo
 VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCallbacks *pAllocator,
                                               VkInstance *pInstance) {
     // This needs to be created before LogMessage is called, as it is dependent on these settings
+    if (layer_settings) {
```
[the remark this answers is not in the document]
mark-lunarg: Multiple instances would likely cause problems, but there was already a single global instance so this isn't any worse than that.  I'm also guessing that multiple instances would be fine, _unless_ they were created with different settings.

layer/tests/tests_promoted.cpp
mark-lunarg: Yes!  All function names and class names should be UpperCamelCase.  Sure, I'll be glad to change these!

layer/tests/tests_capabilities.cpp
mark-lunarg: Yes, will do.

christophe-lunarg: How is this working? 

The previous design aimed at allocating only once the "test infrastructure" to ensure the excuting time remains small.

Is `profiles` allocated only once for this file?

christophe-lunarg (commented):
I don't know that usage of Google Test... If the execution time of the tests remains equivalent to what it was, then this looks like a more clean way to factorize test framework initialization.

mark-lunarg:
Charles' suggestion led me to a GTest-supported method for doing once-per-testsuite initialization and teardown.  I've implemented that for `tests_promoted` and `tests_capabilities` (didn't affect the other tests), and this seems to check all the boxes.  PTAL, I plan to start implementing this for reals tomorrow morning unless there are new things that need to be addressed.

mark-lunarg:
Closing, will finalize and resubmit.

PR #152, opened 2022-02-23
profiles layer: Fix CreateInstance

ziga-lunarg:


mark-lunarg (approved):
Tested, fixes Android crashes.  Thanks Ziga!

PR #155, opened 2022-02-24
script: Allow array sizes to be from constants

charles-lunarg:
Many static sized arrays use an enum value for the size, like
VK_MAX_EXTENSION_NAME_SIZE. The script will now look at these constants for the
size of the array.

This commit was found when trying to generate a vulkan_profiles.cpp that contained a pipelineCacheUUID.

### DIFF
--- a/scripts/genvp.py
+++ b/scripts/genvp.py
@@ -1360,12 +1360,12 @@ class VulkanRegistry():
         xml = etree.parse(registryFile)
         self.parsePlatformInfo(xml)
         self.parseVersionInfo(xml)
+        self.parseConstants(xml)
         self.parseExtensionInfo(xml)
         self.parseStructInfo(xml)
         self.parsePrerequisites(xml)
         self.parseEnums(xml)
         self.parseBitmasks(xml)
-        self.parseConstants(xml)
         self.parseAliases(xml)
         self.parseExternalTypes(xml)
         self.parseFeatures(xml)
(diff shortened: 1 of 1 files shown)


christophe-lunarg (approved):

scripts/genvp.py
```diff
@@ -1453,6 +1453,8 @@ def parseStructInfo(self, xml):
                         elif tail == '[' and enum != None and enum.tail == ']':
                             # [<enum>] case
                             structDef.members[name].arraySize = enum.text
+                            if structDef.members[name].arraySize in self.constants:
+                                structDef.members[name].arraySize = int(self.constants[structDef.members[name].arraySize])
```
aqnuep, line 1457: This is not how this was originally meant to be handled. There's a method in `VulkanRegistry` called `evalArraySize` which is used to map the constant values to their actual integer value.

Probably some call to `evalArraySize` is missing somewhere if this caused issues.
I'll try it on my end what goes wrong with a pipelineCacheUUID.

aqnuep: Please find the suggested fix in PR #158.

I'm not saying that this PR does not fix the same issue, but it does introduce some inconsistency in array size handling.
The original bug was simply a result of me forgetting to update existing code when adding support for dynamic array size evaluation.

charles-lunarg: I'd rather take your PR since its more idiomatic anyhow.
I just did this so I could verify that the profile vulkaninfo created would successfully run through the gen script (it does, with this fix in place).

PR #158, opened 2022-02-25
Fix const array size handling in lambda generators

aqnuep:


charles-lunarg (approved):
Changes look good.
At first I thought the arraySize needed to be converted to a plain int, so pipelineCacheUUID's length would need to turn into 16, but then I remembered that the C/C++ generated code can use those constants just fine.

PR #160, opened 2022-02-25
Add layer test framework, Android layer tests

mark-lunarg:
Added a test framework to the layer tests, factoring out the `main()` functions from the test files.

Creates individual test executables for Win/Lin and a single test executable for Android packaged as **VulkanProfilesLayerTests.apk**.  Updated BUILD.md. 
 
Slight change to library test package generation to prevent namespace collisions during CI build.

@lunarpapillo, the APK ends up in the `apk/out` directory just as the currently library tests do, hopefully this should be sufficient to get CI running.


layer/tests/layer_tests_main.cpp
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    return args;
+}
+
+void addFullTestCommentIfPresent(const ::testing::TestInfo &test_info, std::string &error_message) {
+    const char *const type_param = test_info.type_param();
+    const char *const value_param = test_info.value_param();
+
+    if (type_param != NULL || value_param != NULL) {
```
ziga-lunarg, line 111: Can we use `nullptr` instead of `NULL` and elsewhere too?

layer/profiles.cpp
```diff
@@ -468,7 +469,8 @@ static std::string GetFormatFeature2String(VkFormatFeatureFlagBits2 flags) {
     return result;
 }
 
-static VkProfileLayerSettingsEXT *layer_settings;
+static VkProfileLayerSettingsEXT *layer_settings = new VkProfileLayerSettingsEXT{};
```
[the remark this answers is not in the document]
mark-lunarg: See line 7795.

layer/tests/layer_tests_main.cpp
mark-lunarg: Absolutely, will do.

ziga-lunarg (approved):

PR #171, opened 2022-03-22
Use Vulkan Header 208

christophe-lunarg:


christophe-lunarg:
@charles-lunarg It looks like the Vulkan Loader has a bug that was fixed in sdk-1.3.204 branch but that wasn't brought back to master branch which result in building with v1.3.204 to 1.3.208 tags failing.

With sdk-1.3.204 (OK): http://erusea:8080/job/VulkanProfiles/1670/
With v1.3.208 (FAIL): http://erusea:8080/job/VulkanProfiles/1669/

PR #172, opened 2022-03-22
[WIP] schema: refactoring to improve checking

christophe-lunarg:
Done:
- This refactoring allows not defining a structure without the version or extension related to this structure.
- Add VK_KHR_surface query, this would not be taken into account by the API library but could be implemented by the layer (I think)

TODO:
- Investigating whether we can prevent using an extension when requiring a version which made this extension a core capability.
- Handling partial vs full profiles

### DIFF
--- a/profiles/VP_LUNARG_desktop_portability_2021.json
+++ b/profiles/VP_LUNARG_desktop_portability_2021.json
@@ -1,13 +1,11 @@
 {
-    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.0-154.json#",
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.9.0-154.json#",
     "capabilities": {
         "vulkan11requirements": {
-            "features": {
+            "VK_VERSION_1_1": {
                 "VkPhysicalDeviceMultiviewFeatures": {
                     "multiview": true
-                }
-            },
-            "properties": {
+                },
                 "VkPhysicalDeviceMultiviewProperties": {
                     "maxMultiviewViewCount": 6,
                     "maxMultiviewInstanceIndex": 134217727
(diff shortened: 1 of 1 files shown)


christophe-lunarg:
Done:
- This refactoring allows no defining a structure without the version or extension related to this extension.
- Add VK_KHR_surface query, this would not be taken into account by the API library but could be implemented by the layer (I think)

TODO:
- Investigating whether we can prevent using an extension when requiring a version which made this extension a core capability.
  - Eg: VK_KHR_device_group was propoted to Vulkan 1.1
- Add extensions and version requirements checking.
  - Eg: VK_KHR_device_group requires VK_KHR_device_group_creation

Schema Implementation:
```
{
	"type": "object",
    "additionalProperties" : false,
	"properties": {
    	"VK_KHR_device_group": { 
			"type": "object" 
		},
		"VK_KHR_device_group_creation": { 
			"type": "object" 
		}
	},
	"dependentRequired": {
		"VK_KHR_device_group": ["VK_KHR_device_group_creation"]
	}
}
```
JSON file example (if `VK_KHR_device_group_creation` is removed, then we get an error):
```
{
    "VK_KHR_device_group": {
      
    },
    "VK_KHR_device_group_creation": {
      
    }  
}
```

christophe-lunarg:
We could implement exclusion of extension vs core version like this:
schema:
```
{
	"type": "object",
	"additionalProperties": {
		"type": "object",
		"oneOf": [
			{
				"additionalProperties": false,
				"properties": {
					"VK_VERSION_1_1": {
						"type": "object"
					}
				}
			},
			{
				"additionalProperties": false,
				"properties": {
					"VK_KHR_device_group": {
						"type": "object"
					},
					"VK_KHR_device_group_creation": {
						"type": "object"
					}
				},
				"dependentRequired": {
					"VK_KHR_device_group": [ "VK_KHR_device_group_creation" ]
				}
			}
		]
	}
}
```
JSON files (if VK_VERSION_1_1 is defined, we get errors):
```
{
  	"baseline": {
		"VK_KHR_device_group": {
      
    	},
    	"VK_KHR_device_group_creation": {
          
        }      
    }  
}
```

This said, a device profiles will have both the core and the extension defined so we should somehow make this a mode.

ziga-lunarg (commented):
Looks good to me. If I see this correctly this would also fix the problem where we had structs from extensions used, but the extension itself was not listed, correct?

christophe-lunarg:
> Looks good to me. If I see this correctly this would also fix the problem where we had structs from extensions used, but the extension itself was not listed, correct?

Yes, and ensure the extension dependencies are resolved.

christophe-lunarg:
Similarly we could prevent a profile file to specify both the bundle and the core structure with `oneOf`.

oknath (approved):

Issue #167, opened 2022-03-16
Build of Vulkan-Profiles fails on Fedora 34


davidlunarg:
Trying to build the Vulkan SDK using "vulkansdk all"  on Fedora 34. Get build errors when building Vulkan-Profiles:
```
[ 66%] Building CXX object layer/CMakeFiles/VkLayer_khronos_profiles.dir/vk_layer_table.cpp.o
[ 83%] Building CXX object layer/CMakeFiles/VkLayer_khronos_profiles.dir/__/external/generated-jsoncpp/jsoncpp.cpp.o
[ 83%] Building CXX object layer/CMakeFiles/VkLayer_khronos_profiles.dir/profiles.cpp.o
/home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/jsoncpp.cpp:990:34: error: ‘Reader’ is deprecated: Use CharReader and CharReaderBuilder instead. [-Werror=deprecated-declarations]
  990 | Reader::Char Reader::getNextChar() {
      |                                  ^
In file included from /home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/jsoncpp.cpp:76:
/home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/json/json.h:1511:63: note: declared here
 1511 |     "Use CharReader and CharReaderBuilder instead.") JSON_API Reader {
      |                                                               ^~~~~~
/home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/jsoncpp.cpp:1044:21: error: ‘Reader’ is deprecated: Use CharReader and CharReaderBuilder instead. [-Werror=deprecated-declarations]
 1044 | std::vector<Reader::StructuredError> Reader::getStructuredErrors() const {
      |                     ^~~~~~~~~~~~~~~
In file included from /home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/jsoncpp.cpp:76:
/home/davidpinedo/sdk/1.3.204.1/source/Vulkan-Profiles/external/generated-jsoncpp/json/json.h:1511:63: note: declared here
 1511 |     "Use CharReader and CharReaderBuilder instead.") JSON_API Reader {
      |                                                               ^~~~~~
cc1plus: all warnings being treated as errors
[...]

PR #173, opened 2022-03-24
issue 167: add REGENERATE_LAYER and FATAL_WARNINGS switches

lunarpapillo:
Addressing issue:

    Build of Vulkan-Profiles fails on Fedora 34
    https://github.com/KhronosGroup/Vulkan-Profiles/issues/167

There are two issues here:

-   Building the Vulkan-Profiles layer requires the Python jsonschema module.

-   Fedora compilers give more warnings that Ubuntu compilers.

The build works if the Python "jsonschema" module is added to the
building system, and if -Werror is removed from layers/CMakeLists.txt.

To avoid requiring users to install "jsonschema", a new switch
REGENERATE_LAYER is provided.  If ON (the default), the layer code
for profiles.cpp is generated normally via the gen_layer.py script.
If OFF, the committed repository code for profiles.cpp is used directly
without first regenerating.  The "vulkansdk" script and the SDK
builds will explicitly turn this OFF.

To handle differing warnings between compilers, a new switch
FATAL_WARNINGS is provided; if ON (the default) all compile warnings
in the project will become errors.  If OFF, compile warnings are
issued but will not block the build.  The "vulkansdk" script will
explicitly turns this OFF.

These changes include additional fixes for existing code that provoked
compiler warnings:

- layer-utils/vk_layer_settings.cpp: add cast to avoid:
     warning C4244: '=': conversion from 'int' to 'char', possible loss of data

and a lot of Windows compiler instructions to ignore particular warnings.
(These warnings should be fixed as a later effort.)

### DIFF
--- a/layer/CMakeLists.txt
+++ b/layer/CMakeLists.txt
@@ -69,25 +69,26 @@ if (WIN32)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
     set (CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
 else()
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wpointer-arith -Werror")
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wpointer-arith -Werror")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wpointer-arith")
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wpointer-arith")
 endif()
-
-set(LAYER_PYTHON_FILES ${CMAKE_SOURCE_DIR}/scripts/gen_layer.py) 
-source_group("Python Files" FILES ${LAYER_PYTHON_FILES})
-
-add_custom_target(VpLayer_generate ALL
-    COMMAND ${PYTHON_EXECUTABLE} ${LAYER_PYTHON_FILES} 
-        -registry ${vulkan-headers_SOURCE_DIR}/registry/vk.xml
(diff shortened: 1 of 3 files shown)


ziga-lunarg (approved):

layer/CMakeLists.txt
```diff
@@ -69,25 +69,26 @@ if (WIN32)
...
-        -outLayer ${CMAKE_SOURCE_DIR}/layer/profiles.cpp
-    VERBATIM
-    SOURCES ${LAYER_PYTHON_FILES}
-    DEPENDS ${vulkan-headers_SOURCE_DIR}/registry/vk.xml)
-set_target_properties(VpLayer_generate PROPERTIES FOLDER "Profiles layer")
-
 add_vk_layer(${TARGET_NAME} profiles.cpp vk_layer_table.cpp ${jsoncpp_generated_source} ${jsoncpp_generated_header})
 
-add_dependencies(VkLayer_${TARGET_NAME} VpLayer_generate)
+if(REGENERATE_LAYER)
```
christophe-lunarg, line 78: So it's not regenerated systematically?

lunarpapillo: It is regenerated systematically if `REGENERATE_LAYER=ON` (which is the default).  Regeneration is skipped if `REGENERATE_LAYER=OFF`; in this case the `profiles.cpp` that is checked into the repository is used directly.

lunarpapillo:
> Builds on Fedora 34. However, I do get several warnings. Not sure if these are new, or were there already. Build output is attached.

I see a bunch of CMake messages, but no compile warnings, in your posted output - was there more that you didn't capture?

I presume you had modified `vulkansdk` by hand to pass `REGENERATE_LAYER=OFF` and `FATAL_WARNINGS=OFF`...?

lunarpapillo:
The corresponding **Vulkan-SDK-Packaging** change is here: https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/merge_requests/239

PR #181, opened 2022-04-06
Update to 211 header

ziga-lunarg:


christophe-lunarg (approved):
So effectively, we need a PR for 210 and 211 with tags for each so that we can generate the schema for each Vulkan Header revision.

christophe-lunarg:
Ah oups, I already pushed 210 previously so it was all perfect!

PR #182, opened 2022-04-08
script: fix unused variable warnings

TheLavaBlock:
if you have set warnings as errors, it causes problems


ziga-lunarg:
@TheLavaBlock thanks for the fix. You should also build the layer with these changes and include them in the PR.

TheLavaBlock:
Here we go...

btw: [BUILD.md](https://github.com/KhronosGroup/Vulkan-Profiles/blob/master/BUILD.md#linux-and-macos-build) is not up-to-date and unfortunately also contains error :eyes:

ziga-lunarg:
> Here we go...
> 
> btw: [BUILD.md](https://github.com/KhronosGroup/Vulkan-Profiles/blob/master/BUILD.md#linux-and-macos-build) is not up-to-date and unfortunately also contains error 👀

Do you see anything other than `cd VulkanTools` instead of `cd Vulkan-Profiles`?

TheLavaBlock:
yeah that's the error and I think `git@github.com:LunarG/VulkanProfiles.git` is no longer up to date, right?

ziga-lunarg:
Will be fixed in #183 , thanks!

PR #186, opened 2022-04-15
profiels: Update schemas in sample configs

ziga-lunarg:


ziga-lunarg:
Looks like this has already been done

PR #197, opened 2022-05-07
profiles: Implement bits minmul and minpot limit types

### LINKED ISSUE #192: Implement new `bits`, `minpot` and `minmul` limittypes
`bits`, `minpot` and `minmul` are new limittypes we are introducing to improve the checking of some capabilities and creating a tool to combine profiles.

`bits` behaves like `max` for the toolset except that is used so that the Vulkan developers know they can use samller values while `bits` is the value effectively used by the driver and hardware. This said, contrary to other "behavior" capabilities, this one can be emulate to less bits.

`minpot` behaves like the `min` limittype but the value also need to be a power of two value. It's typically used for alignment values and will be particularly useful to create a tool to combine profiles.

`minmul` behaves like the `min` limittype but the value also need to be a multiple of the set value. It's typically used for `granularity` values. When combining two profiles we can compute the new value using Profile A values * Profile B Values so that the value is a multiple of both.

The values that should be `minpot` and `minmul` are documented by this PR: https://github.com/KhronosGroup/Vulkan-Profiles/pull/193

ziga-lunarg:
Closes #192


scripts/genvp.py
```diff
@@ -2281,9 +2281,18 @@ def gen_structCompare(self, fmt, structDef, var, values, parentLimittype = None)
                 elif limittype == 'max':
                     # Compare max limit by checking if device value is greater than or equal to profile value
                     comparePredFmt = '{0} >= {1}'
+                elif limittype == 'bits':
+                    # Behaves like max, but smaller values are allowed, so no check needed
+                    comparePredFmt = None
```
christophe-lunarg, line 2286: I am not sure this is quite right. `comparePredFmt = '{0} >= {1}'` was probably right here.

christophe-lunarg (approved):

scripts/genvp.py
ziga-lunarg: But aren't smaller values allowed?

christophe-lunarg:
I merged the PR because it's fine but it's missing some limittypes (behavior et mulpot) that are documented on the Gitlab issue:
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5172

PR #228, opened 2022-07-14
profiles: Update to 221 header

ziga-lunarg:


scripts/genvp.py
```diff
@@ -2040,6 +2040,12 @@ def applyWorkarounds(self):
             self.structs['VkPhysicalDeviceMeshShaderPropertiesNV'].members['meshOutputPerVertexGranularity'].limittype = 'min,mul' # should be 'min,mul'
             self.structs['VkPhysicalDeviceMeshShaderPropertiesNV'].members['meshOutputPerPrimitiveGranularity'].limittype = 'min,mul' # should be 'min,mul'
 
+        if 'VkPhysicalDevicePipelineRobustnessPropertiesEXT' in self.structs:
```
ziga-lunarg, line 2043: @christophe-lunarg Had to add this to the workaround list, can you verify it is correct?

christophe-lunarg: This looks good!

PR #258, opened 2022-08-10
profile: add all core extensions in Roadmap 2022

christophe-lunarg:


r-potter:
@christophe-lunarg I just saw the notification for this go past in my email. Is this requiring that an implementation expose the full list of extensions via `vkEnumerate*ExtensionProperties` in addition to exposing them via the 1.3 core, or am I misunderstanding the PR? If I've misunderstood then please ignore this.

If so, this seems incorrect. It's valid for implementations to expose functionality via the core API without exposing the original extension, and Roadmap 2022 does not require support for the original extension. Most implementations will expose both, but it's not a requirement of the specification.

This is especially true for the extensions that add image formats (`VK_EXT_4444_formats`, `VK_EXT_ycbcr_2plane_444_formats`), where the requirements for the core API differ from the extensions (i.e. in core you don't actually have to support the formats, just recognise the format enums).

christophe-lunarg:
@r-potter then I misunderstood how things were supposed to work. Thanks for the clarification, I'll revert this.

PR #282, opened 2022-09-08
VP_KHR_roadmap_2022: Reduce compute workgroup dims

gnl21:
The Vulkan spec says that the limit for maxComputeWorkGroupSize in the roadmap_2022 profile should be [256, 256, 64], so fix this in the profile spec.


christophe-lunarg:
This PR is correct: https://registry.khronos.org/vulkan/specs/1.3-khr-extensions/html/vkspec.html#roadmap-2022

PR #284, opened 2022-09-08
profiles: Update to 228 header

ziga-lunarg:
- Update known good


ziga-lunarg:
Made a typo in the commit, the header is 227

PR #305, opened 2022-10-18
profile: on Linux depends on tinyxml2-dev package

christophe-lunarg:


lunarpapillo:
This won't work correctly on **SDK** builds, as it places a quiet dependency on the version of `tinyxml` that happens to be installed on a build machine.

Should we ever need to reproduce a historical SDK, we need the ability to use *exactly* the same components of all dependencies, including `tinyxml`.  Our standard procedure for encoding these dependencies is to put them all on `config.json`, which we guarantee will have everything a user needs to recreate a build.

The `tinyxml` dependency should be in `config.json`.  The **Vulkan-Profiles** build should accept a CMake variable `TINYXML_INSTALL_DIR` and, if such is set, `tinyxml` will be used from that directory instead of being fetched by `FetchContent()`.

Alternatively, the required source from `tinyxml` could be checked directly into the **Vulkan-Profiles** repository, like `jsoncpp` is.  But that brings its own issues, of course.

lunarpapillo (changes requested):
A quiet dependency on the version of `tinyxml` that happens to be installed on a build machine prevents us from correctly rebuilding an older SDK.

We've not (yet) had to do that, but we're supposed to maintain the ability.

christophe-lunarg:
I agree, this is not the right approach, I need to figure out what's going on in the initial issue.

PR #309, opened 2022-10-26
library: Add support for enabling multiple profiles when creating Vulkan instances

christophe-lunarg:
This PR is breaking Profiles API library backward compatible


christophe-lunarg:
Ending experiment.

PR #320, opened 2022-11-07
Updated to more secure vsnprintf to fix deprecation errors with Apple…

richard-lunarg:
The latest compiler update from Apple is now erroring out on these deprecated and "unsafe" versions of vsprintf. This commit replaces them with vsnprintf, and specifies the sizes of the buffers. Confirmed this now build with the latest Apple toolset.


christophe-lunarg (approved):
Looks good except that somehow all the C.I. build on Windows and Linux got aborted...

Issue #337, opened 2023-01-09
SIMULATE_FORMATS_BIT with VK_KHR_format_feature_flags2


spencer-lunarg:
If my `profile.json` file looks like

```json
"VK_FORMAT_B8G8R8A8_UNORM": {
                    "VkFormatProperties": {
                        "linearTilingFeatures": [
                            "VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT",
                        ],
                        "optimalTilingFeatures": [
                            "VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT",
                        ],
                        "bufferFeatures": []
                    }
                },
 ```
 
 but my profile also supports Vulkan 1.3 or `VK_KHR_format_feature_flags2` (which is an unique extension because if it is exposed, it has to be supported, no enabling needed)
 
 I am finding that calling 
 
 ```cpp
VkFormatProperties3KHR fmt_props_3;
VkFormatProperties2 fmt_props_2;
fmt_props_2.pNext = &fmt_props_3;
vkGetPhysicalDeviceFormatProperties2(physical_device, format, &fmt_props_2);
```

Will not return `VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_BIT` like I would expect it too

-----

I see 2 ways around this

1. Force everyone to duplicate both `VkFormatProperties` and `VkFormatProperties3`, and have tooling to validate when it is forgotten
2. Have `JsonLoader::GetFormat` have some logic to make sure `profile_properties_3` and `profile_properties` have the same union

I like number 2, there is already CTS test to make sure all implementation have the same union of formats for both

> The bits reported in **linearTilingFeatures**, **optimalTilingFeatures** and **bufferFeatures** must include the bits reported in the corresponding fields of **VkFormatProperties2::formatProperties**.

spencer-lunarg:
for sake of unblocking anyone else with this issue by chance, this script will duplicate it over 

```python
import json

INPUT = 'profile.json'
OUTPUT = 'new_profile.json'

with open(INPUT, 'r') as file:
    data = json.load(file)

oldFormats = data['capabilities']['device']['formats']
newFormats = {}

for format in oldFormats:
    newFormats[format] = oldFormats[format]
    newFormats[format]['VkFormatProperties3'] = dict()
    newFormats[format]['VkFormatProperties3']['linearTilingFeatures'] = []
[...]

PR #342, opened 2023-01-16
Improve handling of  by profiles layer #337

christophe-lunarg:


scripts/gen_profiles_layer.py
```diff
@@ -1251,39 +1251,44 @@ class JsonLoader {
     const auto &member = formats[format_name];
     for (const auto &name : member.getMemberNames()) {
         const auto &props = member[name];
-        if (name == "VkFormatProperties") {
+        if (name == "VkFormatProperties" || name == "VkFormatProperties3" || name == "VkFormatProperties3KHR") {
             for (const auto &feature : props["linearTilingFeatures"]) {
                 profile_properties.linearTilingFeatures |= StringToVkFormatFeatureFlags(feature.asString());
+                profile_properties_3.linearTilingFeatures |= StringToVkFormatFeatureFlags2(feature.asString());
```
ziga-lunarg, line 1257: Why is this combined? And is it correct? You are passing the same `feature.asString()` string to different functions that expect different values. One has `VkFormatFeatureFlags` and the other `VkFormatFeatureFlags2`

christophe-lunarg: `VkFormatFeatureFlags2` maps to `VkFormatFeatureFlags` except there are more values.

Basically, the code will return nothing when it doesn't find the string.

PR #346, opened 2023-01-21
profiles: Update to 239 header

ziga-lunarg:


christophe-lunarg:
@ziga-lunarg  tests_generated.cpp is not longer checked in as it's a generated file.

PR #349, opened 2023-01-23
fix usage of deprecated header

christophe-lunarg:


layer-utils/vk_layer_settings.cpp
```diff
@@ -40,7 +40,6 @@
 #include <vulkan/vk_layer.h>
 // sdk_platform header redefines NOMINMAX
 #undef NOMINMAX
-#include <vulkan/vk_sdk_platform.h>
```
juan-lunarg, line 43: You should remove lines 41/42 as well.

This header file was deprecated for the 1.3.239 release due to https://github.com/KhronosGroup/Vulkan-Headers/issues/316

PR #350, opened 2023-01-23
profiles: Remove reference to vk_sdk_platform.h

juan-lunarg:


juan-lunarg:
Leftover from https://github.com/KhronosGroup/Vulkan-Profiles/pull/349

PR #357, opened 2023-02-16
CMake cleanup

juan-lunarg:


CMakeLists.txt
```diff
@@ -5,22 +5,16 @@ enable_testing()
 
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "Minimum OS X deployment version")
 
-project(VULKAN_PROFILES LANGUAGES CXX C)
+project(VULKAN_PROFILES LANGUAGES CXX)
```
juan-lunarg, line 8: It's best practice to minimize the compilers used for your project. It complicates the CMake code.

CMakeLists.txt
```diff
@@ -5,22 +5,16 @@ enable_testing()
...
 # Regenerating profiles requires Python 3.7.2 or better (for typing.OrderedDict).
 # Regeneration jsoncpp requires Python 3.
 # Otherwise Python is not required, as checked-in generated code and
 # headers will be used.
 option(PROFILES_BUILD_TESTS "Build profile tests" ON)
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" CACHE PATH "Archive output dir.")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" CACHE PATH "Library output dir.")
-set(CMAKE_PDB_OUTPUT_DIRECTORY     "${CMAKE_BINARY_DIR}/bin" CACHE PATH "PDB (MSVC debug symbol)output dir.")
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin" CACHE PATH "Executable/dll output dir.")
```
juan-lunarg, line 19: This was needed for the jsoncpp but isn't actually good practice. Now that jsoncpp is built standalone and consumed via find_package we don't need to do this anymore.

CMakeLists.txt
```diff
@@ -5,22 +5,16 @@ enable_testing()
...
-
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
-set(CMAKE_C_VISIBILITY_PRESET "hidden")
```
juan-lunarg, line 23: Now that we only specify CXX for the language we can remove CMake code like this.

CMakeLists.txt
```diff
@@ -139,11 +133,6 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # Define GNU standard installation directories.
 include(GNUInstallDirs)
 
-if(WIN32 AND CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-    # Windows: if install locations not set by user, set install prefix to "<build_dir>\install".
-    set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "default install path" FORCE)
-endif()
-
```
juan-lunarg, line 146: This was removed from vvl/loader a bit ago. It's not needed and can cause users issues.

CMakeLists.txt
```diff
@@ -157,29 +146,24 @@ if (BUILD_WERROR)
 endif()
 
 # Platform-specific compiler switches
-if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
-    add_compile_options(-Werror)
-endif()
-if(${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
-    add_compile_options(-Werror)
```
juan-lunarg, line 164: BUILD_WERROR handles this.

layer/CMakeLists.txt
```diff
@@ -45,11 +45,10 @@ add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
 set(TARGET_NAME khronos_profiles)
 
 if (WIN32)
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
-    set (CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
+    add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
+    add_compile_options(/bigobj)
 else()
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wpointer-arith")
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wpointer-arith")
```
juan-lunarg, line 52: 2 things.

1.) CMAKE_C_FLAGS isn't needed at all anymore.
2.) Unless you HAVE to don't touch these variables.

CMakeLists.txt
juan-lunarg: This is one of the benefits of using find_package versus add_subdirectory.

add_subdirectory makes you have to care about the quality of the CMake code of the libraries you consume. Versus find_package just uses the already built artifacts.

CMakeLists.txt
juan-lunarg: Also CMAKE_C_COMPILER_ID doesn't exist anymore. Now it's just CMAKE_CXX_COMPILER_ID

CMakeLists.txt
ziga-lunarg: What is the default without this?

scripts/known_good.cmake
```diff
@@ -1,12 +0,0 @@
...
-set(GIT_VULKAN_HEADER_REPOSITORY https://github.com/KhronosGroup/Vulkan-Headers)
-set(GIT_VULKAN_HEADER_COMMIT v1.3.239)
-
-# Vulkan Loader
-set(GIT_VULKAN_LOADER_REPOSITORY https://github.com/KhronosGroup/Vulkan-Loader)
-set(GIT_VULKAN_LOADER_COMMIT v1.3.239)
-
-# Vulkan Validation Layer
-set(GIT_VULKAN_VALIDATION_REPOSITORY https://github.com/KhronosGroup/Vulkan-ValidationLayers)
-set(GIT_VULKAN_VALIDATION_COMMIT v1.3.239)
```
ziga-lunarg, line 11: Where will this be set now?

CMakeLists.txt
juan-lunarg: Not sure. The main issue is it prevents users/ci from setting [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/latest/envvar/CMAKE_PREFIX_PATH.html) as an environment variable.

scripts/known_good.cmake
juan-lunarg: This is set in the known_good.json

ziga-lunarg (approved):
Good cleanup and helpful comments

CMakeLists.txt
juan-lunarg: Also it breaks the new `--install` functionality which is also quite useful for package managers (and local users)

```
cmake --install build --prefix build/install
```

CMakeLists.txt
juan-lunarg: Looks like profiles maybe depending on the layout of the build directory so I reverted this change for now.

juan-lunarg:
Looks like the Android CI is having some issues unrelated to this PR

juan-lunarg:
Android CI issues

lunarpapillo:
?  Was this cherry-picked, or abandoned?

juan-lunarg:
Abandoned for now

PR #361, opened 2023-02-20
Resolving SDK build

christophe-lunarg:


juan-lunarg (changes requested):
include_directories applies this on a global scope.

Use target_include_directories

christophe-lunarg:
This is what the repository already used but the SDK fail to build and find the validation layer headers.

> include_directories applies this on a global scope.
> 
> Use target_include_directories

PR #364, opened 2023-02-20
ci: Test MacOS build

juan-lunarg:


juan-lunarg:
Tested in my fork:
https://github.com/juan-lunarg/Vulkan-Profiles/actions/runs/4226455029

CMakeLists.txt
```diff
@@ -48,7 +46,7 @@ if (UPDATE_DEPS)
     message("********************************************************************************")
 
     set(optional_args)
-    if (NOT BUILD_TESTS)
```
christophe-lunarg, line 51: Is this actually a good things this variable is not name BUILT_TESTS? Isn't it inconsistent with other repository?

christophe-lunarg:
If we could have macOS C.I., that would be nice!

CMakeLists.txt
juan-lunarg: It is inconsistent. But in this case I'd argue it's a good thing. It's generally good practice to group your cache variables together so it shows up nicely in cmake-gui.

image

juan-lunarg: Main reason I haven't fixed it in vvl is because it's already established.

PR #367, opened 2023-02-21
vku cleanup

### LINKED ISSUE #366: Do we need to install vku?
https://github.com/KhronosGroup/Vulkan-Profiles/blob/c5334c4a776aab9aca82a81c4dc312840086b9f0/layer-utils/CMakeLists.txt#L29

Why is vku being installed? Is it a deliverable of Vulkan-Profiles?

juan-lunarg:
closes #366


christophe-lunarg (approved):
If that doesn't break anything, great!

PR #368, opened 2023-02-21
cmake: Don't specify *_OUTPUT_DIRECTORY

juan-lunarg:


juan-lunarg:
Profiles currently relies on specifying the *_OUTPUT_DIRECTORY for testing.

PR #371, opened 2023-02-22
Update to header 241

ziga-lunarg:


scripts/gen_profiles_layer.py
```diff
@@ -3315,31 +3315,31 @@ class JsonLoader {
 // Function symbols statically exported by this layer's library //////////////////////////////////////////////////////////////////
 // Keep synchronized with VisualStudio's VkLayer_khronos_profiles.def
 
-VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(VkInstance instance, const char *pName) {
+VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(VkInstance instance, const char *pName) {
```
juan-lunarg, line 3318: Vulkan layers need to export function.

See how VVL does it here:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/de852e626f2f6e3027448eb5c2af10af26d197e1/layers/generated/chassis.cpp#L15868

See this VVL PR for full context:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/be666f39c462905ef245d88ff8f36bc8a8e7cf8e

juan-lunarg (dismissed):
~~Tests are failing because critical functions aren't being exported.~~ Looks like it's not just the function not being exported.

juan-lunarg (approved):

scripts/gen_profiles_layer.py
```diff
@@ -2791,7 +2791,7 @@ class JsonLoader {
     uint32_t pCount_copy = *pCount;
 
     PhysicalDeviceData *pdd = PhysicalDeviceData::Find(physicalDevice);
-    const uint32_t src_count = pdd ? static_cast<uint32_t>(pdd->simulation_extensions_.size()) : 0;
+    //const uint32_t src_count = pdd ? static_cast<uint32_t>(pdd->simulation_extensions_.size()) : 0;
```
christophe-lunarg, line 2794: Is this because it's dead code?

scripts/gen_profiles_layer.py
```diff
@@ -4394,7 +4407,7 @@ def generate_string_to_format(self, formats):
 
     args = parser.parse_args()
 
-    registryPath  = './build/_deps/vulkan-headers-src/registry/vk.xml'
+    registryPath  = './external/Vulkan-Headers/build/install/share/vulkan/registry/vk.xml'
```
christophe-lunarg, line 4410: Is that actually correct ? @juan-lunarg 
Actually, how could it even work without this... Oo

christophe-lunarg: I think this default path is only when working locally on the scripts so it's fine.

christophe-lunarg (approved):
Nice work!

scripts/gen_profiles_layer.py
juan-lunarg: Not a very clean approach TBH.

VVL passes the registry path to our scripts like so:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/f83f141f879d1d148a1c9b1026a9c724d4ad2bcf/layers/CMakeLists.txt#L157

See Vulkan-Headers doc:
https://github.com/KhronosGroup/Vulkan-Headers/blob/main/BUILD.md

juan-lunarg: However, this approach is consistent with what was already being done. So it looks good to me.

scripts/gen_profiles_layer.py
ziga-lunarg: Yes, variable was unused, the code where it was used before is also commented out

scripts/gen_profiles_layer.py
ziga-lunarg: The `_deps` folder doesn't exist anymore so I just updated this

PR #375, opened 2023-03-07
Update CMake minimum to 3.17

juan-lunarg:
- Raise minimum to 3.17.2
- Explicilty test 3.17.2 via github actions similar to the vulkan validation layers

### DIFF
--- a/.github/workflows/profiles.yml
+++ b/.github/workflows/profiles.yml
@@ -51,6 +51,31 @@ jobs:
       - name: Install
         run: cmake --install build --prefix build/install --config ${{matrix.config}}
 
+  ubuntu-latest:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        config: [Debug, Release]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Install jsonschema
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14.0)
+cmake_minimum_required(VERSION 3.17.2)
```
juan-lunarg, line 1: The intent is to match the vulkan validation layers and reduce friction for developers who work on multiple CMake repos.

juan-lunarg: Also the vulkan loader, vulkan extension layer, etc. also use 3.17.2 as the minimum

.github/workflows/profiles.yml
```diff
@@ -51,6 +51,31 @@ jobs:
...
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Install jsonschema
+        run: python3 -m pip install jsonschema
+      - name: Install Dependencies
+        run: sudo apt update && sudo apt install --yes --no-install-recommends libwayland-dev libxrandr-dev
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.17.2
```
juan-lunarg, line 71: Ensure Vulkan-Profiles builds with exactly `3.17.2`. This ensures stability for system package managers who tend to rely on this functionality.

PR #378, opened 2023-03-14
known_good.json optimization & cleaning up update_deps.py

juan-lunarg:
Optimizes vvl ci by not downloading loader if we don't need to (IE we are not building tests)
Move update deps CMake logic to scripts/CMakeLists.txt (This matches what the loader/vvl currently do)

### DIFF
--- a/.github/workflows/profiles.yml
+++ b/.github/workflows/profiles.yml
@@ -42,7 +42,7 @@ jobs:
         run: python3 -m pip install jsonschema
       - uses: lukka/get-cmake@latest
       - name: Configure/Generate
-        run: cmake -S. -B build -D UPDATE_DEPS=ON -DCMAKE_BUILD_TYPE=${{matrix.config}} -G "Ninja"
+        run: cmake -S. -B build -D PROFILES_BUILD_TESTS=OFF -D UPDATE_DEPS=ON -DCMAKE_BUILD_TYPE=${{matrix.config}} -G "Ninja"
         env:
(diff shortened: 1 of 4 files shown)


ziga-lunarg (approved):
Not that I understand everything, but LGTM

.github/workflows/profiles.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: python3 -m pip install jsonschema
       - uses: lukka/get-cmake@latest
       - name: Configure/Generate
-        run: cmake -S. -B build -D UPDATE_DEPS=ON -DCMAKE_BUILD_TYPE=${{matrix.config}} -G "Ninja"
+        run: cmake -S. -B build -D PROFILES_BUILD_TESTS=OFF -D UPDATE_DEPS=ON -DCMAKE_BUILD_TYPE=${{matrix.config}} -G "Ninja"
```
christophe-lunarg, line 45: @juan-lunarg this is weird to me, we do need to build the profiles tests on C.I.

juan-lunarg: Does testing work on MacOS?

christophe-lunarg:
I am pretty sure @lunarpapillo enabled it rescently.

.github/workflows/profiles.yml
christophe-lunarg: I am pretty sure @lunarpapillo enabled it rescently.

Issue #385, opened 2023-03-23
Windows install target doesn't package everything to the traditional SDK locations


lunarpapillo:
## Configuration Files
The **Windows SDK** has traditionally shipped the **Vulkan-Profiles** configuration files in `Config\VK_LAYER_KHRONOS_profiles\`, e.g.:
```
Config\VK_LAYER_KHRONOS_profiles\VP_ANDROID_baseline_2021.json
Config\VK_LAYER_KHRONOS_profiles\VP_ANDROID_baseline_2022.json
Config\VK_LAYER_KHRONOS_profiles\VP_KHR_roadmap_2022.json
Config\VK_LAYER_KHRONOS_profiles\VP_LUNARG_desktop_baseline_2022.json
Config\VK_LAYER_KHRONOS_profiles\VP_LUNARG_desktop_portability_2022.json
```

But when the `install` target was added for Windows, those files started to appear instead in `share\vulkan\config\VK_LAYER_KHRONOS_PROFILES`.

Either the Windows install target should put these files into a top-level `Config` directory, or the SDK packaging should move them to `share\vulkan\config\.  (The latter could be a breaking change for SDK users, so we should consider this carefully if this is the intended plan.)

## Library Files
The `install` target is also shipping `bin\VkLayer_khronos_profiles.lib`, which provokes issues:
- We've never shipped a layer `.lib` file (for any layer) before.
- The `.lib` files we are shipping are all in the `Lib\` directory.

We should do one of:
- remove the `bin\VkLayer_khronos_profiles.lib` from the `install` target
- move `bin\VkLayer_khronos_profiles.lib` to `lib\VkLayer_khronos_profiles.lib`, and add `.lib` files to the SDK for all the other layers we package

juan-lunarg:
> But when the install target was added for Windows, those files started to appear instead in share\vulkan\config\VK_LAYER_KHRONOS_PROFILES.

I'll defer to @christophe-lunarg

> remove the bin\VkLayer_khronos_profiles.lib from the install target

We should remove the `VkLayer_khronos_profiles.lib` it's an artifact I've been meaning to get to and shouldn't be installed. Similar to the vulkan extension layer.

lunarpapillo:
I found one more with my next SDK build attempt: http://erusea:8080/view/Manual/job/manual-sdk/516/console
```
[...]

PR #386, opened 2023-03-23
cmake: Fix json install

juan-lunarg:
See https://github.com/KhronosGroup/Vulkan-Profiles/issues/385


juan-lunarg:
closing for now. Since it looks like there is multiple install issues.

lunarpapillo (commented):
Perfect, thanks.

PR #397, opened 2023-04-04
layer: add setting to control vulkan feature init

christophe-lunarg:
image
image

image
image


christophe-lunarg:
Investigating "CI VulkanProfiles build # 2457 failed" it's only due to an Android device being unreachable.

PR #399, opened 2023-04-05
cmake: Use VulkanUtilityLibrary

juan-lunarg:


ziga-lunarg (approved):

christophe-lunarg (approved):

christophe-lunarg:
> Note in order to fully remove the dependency on the validation layers it looks like we need to add functionality that covers:
> 
> * vk_layer_dispatch_table.h
> * vk_dispatch_table_helper.h
> * vk_layer_utils.h
> * vk_enum_string_helper.h
> 
> This will be a good first step in removing a dependency on the validation layers.
> 
> Thoughts?

@juan-lunarg agreed!

PR #407, opened 2023-04-14
PDB installation

juan-lunarg:


CMakeLists.txt
```diff
@@ -78,6 +78,14 @@ endif()
 if(${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
     add_compile_options(-Werror)
 elseif(MSVC)
+    # PDBs aren't generated on Release builds by default.
+    add_compile_options("$<$<CONFIG:Release>:/Zi>")
+    add_link_options("$<$<CONFIG:Release>:/DEBUG:FULL>")
```
juan-lunarg, line 83: This will generate a PDB on Release configs

CMakeLists.txt
```diff
@@ -78,6 +78,14 @@ endif()
...
+    # Remove unreferenced code/data.
+    add_link_options("$<$<CONFIG:Release>:/OPT:REF>")
+    # Merge duplicate data/functions
+    add_link_options("$<$<CONFIG:Release>:/OPT:ICF>")
```
juan-lunarg, line 87: This is needed to prevent a performance regression due to usage of `/DEBUG`

When you use `/DEBUG` MSVC disables linker optimizations for some misguided reason.

layer/CMakeLists.txt
```diff
@@ -159,5 +159,8 @@ endif()
 if (WIN32)
     install(FILES ${INTERMEDIATE_FILE} DESTINATION ${LAYER_INSTALL_DIR} RENAME ${OUTPUT_FILE_FINAL_NAME})
 endif()
+if (MSVC)
```
juan-lunarg, line 162: MINGW doesn't generate PDBs so we check for MSVC instead of WIN32

PR #419, opened 2023-05-06
layer: Update to header 250

ziga-lunarg:


juan-lunarg (dismissed):
We should update the known_good.json for the utility libraries first.
https://github.com/KhronosGroup/Vulkan-Utility-Libraries

christophe-lunarg (approved):

christophe-lunarg:
Should we use "v1.3.250" for "commit" of [Vulkan-Utility-Libraries](https://github.com/KhronosGroup/Vulkan-Utility-Libraries)?

PR #439, opened 2023-05-31
profiles: Update utility libraries to v1.3.251

juan-lunarg:


christophe-lunarg:
Actually, I used main on purpose to be able to catch issues in the Utility-Liabaries repository following the spirit of the validation layer using profiles main to catch issues.

juan-lunarg:
> Actually, I used main on purpose to be able to catch issues in the Utility-Liabaries repository following the spirit of the validation layer using profiles main to catch issues.

We use main in `common_ci.py` which only affects testing. `known_good.json` should always produce a stable build.

christophe-lunarg:
Ah ok then.

juan-lunarg:
Also in `common_ci.py` we aren't grabbing dependencies required to build the validation layers. We are grabbing binary dependencies required to enable testing.

PR #449, opened 2023-06-21
cmake: Update update_deps.py files

juan-lunarg:


juan-lunarg:
> CI VulkanProfiles build # 2615 failed.

See https://github.com/LunarG/VulkanTests/pull/395

PR #452, opened 2023-06-21
cmake: Use modern Python3 CMake support

juan-lunarg:
From the 3.27 release notes:
The FindPythonInterp and FindPythonLibs modules, which have been deprecated since CMake 3.12, have been removed by policy CMP0148. Port projects to FindPython3, FindPython2, or FindPython.


juan-lunarg:
Man github actions takes a lot longer to start for profiles than it does for VVL. Does VVL has some sort of higher priority?

PR #469, opened 2023-07-28
profiles: Update to header 260

ziga-lunarg:
There is one issue, the generator will complain that the struct `VkPhysicalDeviceHostImageCopyPropertiesEXT` is ill-formed, because it contains dynamic arrays

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -6,7 +6,7 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.257"
+            "commit": "v1.3.260"
         },
         {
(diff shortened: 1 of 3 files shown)


scripts/known_good.json
```diff
@@ -6,15 +6,15 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.257"
+            "commit": "v1.3.259"
```
christophe-lunarg, line 9: Is this exepected?

christophe-lunarg (approved):
Looks good but it looks like it's Header 259

scripts/known_good.json
ziga-lunarg: Forgot to include this file with the latest push, fixed now

christophe-lunarg:
Looks great!

PR #477, opened 2023-08-14
scripts: Improve handling of API variants

aqnuep:
This PR introduces some improvements to the generator scripts to better handle API variants defined in the vk.xml, as the previous code did not handle some of the aspects of the new vk.xml syntaxes introduced for API variant dependent definitions in a general way.

### DIFF
--- a/scripts/gen_profiles_solution.py
+++ b/scripts/gen_profiles_solution.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 #
 # Copyright (c) 2021-2023 LunarG, Inc.
+# Copyright (c) 2023-2023 RasterGrid Kft.
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 # you may not use this file except in compliance with the License.
@@ -1340,7 +1341,7 @@ class VulkanLimit():
 
 
 class VulkanVersionNumber():
-    def __init__(self, versionStr):
+    def __init__(self, versionStr, targetApi = None, versionName = None):
         match = re.search(r"^([1-9][0-9]*)\.([0-9]+)$", versionStr)
         if match != None:
             # Only major and minor version specified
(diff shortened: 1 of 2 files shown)


christophe-lunarg:
@aqnuep, I think I would feel a little more confident about this if I had a better idea of what specific cases it resolves.

For example, could you had some tests, maybe adding "profiles" here E:\Github\khronos\Vulkan-Profiles - Bugfix\profiles\test\data and manually checking the specific cases in E:\Github\khronos\Vulkan-Profiles - Bugfix\layer\tests\tests_mechanism.cpp

?

scripts/gen_profiles_solution.py
```diff
@@ -1421,7 +1455,7 @@ def __init__(self, xml, upperCaseName):
         self.limits = dict()
         self.platform = xml.get('platform')
         self.provisional = xml.get('provisional')
-        self.promotedTo = xml.get('promotedto')
+        self.promotedTo = xml.get('promotedto').split(',') if xml.get('promotedto') is not None else []
```
christophe-lunarg, line 1458: Is there a case where an extension is promoted to multiple extensions? Or is this for promotedto an extension and a core version maybe?

christophe-lunarg (approved):

christophe-lunarg:
I think the changes are mostly to better handle Vulkan and Vulkan SC it seems reasonnable.

It looks like it also fix some cases regarding promotedto and alias for which I could enjoy seeing additional tests to prevent future changes accidently break these cases.

aqnuep:
Yes, most of these were detected by testing against Vulkan SC content, but are rather generic fixes so far.

The problem is that in order to write tests for this you'd actually have to create custom vk.xml inputs to test with, e.g. one where `supported="vulkan,vulkansc"` is replaced with `supported="vulkansc,vulkan"` in the extension tags, which the code before this PR doesn't handle.

PR #479, opened 2023-08-16
layer: Fix KHR GPDP2 dependency

aqnuep:
The Profiles layer incorrectly depends on the KHR variant of GPDP2 in some places even when Vulkan 1.1 is available.


christophe-lunarg (approved):
That's fine this said the profiles layer is suposed to require Vulkan 1.1.

It's only the profiles library that required Vulkan 1.0 and it's actually something we may be able to evaluate again.

aqnuep:
> That's fine this said the profiles layer is suposed to require Vulkan 1.1.

Well, that's even worse, because before this fix the layer wouldn't work on a Vulkan 1.1 implementation that doesn't also export KHR GPDP2, which is not a requirement :)

PR #489, opened 2023-08-25
profiles: Update to header 262

ziga-lunarg:


scripts/known_good.json
```diff
@@ -6,7 +6,7 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.261"
+            "commit": "v1.3.262"
         },
         {
             "name": "Vulkan-Utility-Libraries",
```
juan-lunarg, line 12: Update the commit for the utility libraries as well. Otherwise it won't get the updated dispatch tables.

juan-lunarg (changes requested):

scripts/known_good.json
juan-lunarg: https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/82

juan-lunarg: @mikes-lunarg something is wrong with the 1.3.262 tag for the utility libraries.

https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commits/v1.3.262

juan-lunarg: @ziga-lunarg the tag has been fixed for the utility libraries.

ziga-lunarg: Doesn't seem to work

christophe-lunarg (approved):

scripts/known_good.json
juan-lunarg: I'll take a look.

juan-lunarg: Okay. Something is wrong with the tag still. I see what you mean. Use `b0712dfdb1f38cf707d0295c9fc687635046c713` instead.

juan-lunarg (approved):

PR #491, opened 2023-08-30
layer: Handle platform protect defines

aqnuep:
This PR enables proper handling of platform protection #defines in the Profiles Layer so that in case any extension specific to a platform introduces new device features or properties, then the layer will continue to build fine per the enabled `VK_USE_PLATFORM_XX` definitions.

At the moment there are no such Vulkan extensions, but that may change in the future.

No similar changes seem to be necessary for the Profiles Library, as the library is generated with specific profiles, and their definitions are only included if all the corresponding `VK_FOO_bar` definitions corresponding to the contained extensions are defined, which are already preconditioned by the corresponding platform protection #defines.

### DIFF
--- a/scripts/gen_profiles_layer.py
+++ b/scripts/gen_profiles_layer.py
@@ -2790,6 +2790,32 @@ class VulkanProfilesLayerGenerator():
             f.write(self.generate_enumerate_physical_device())
             f.write(GET_INSTANCE_PROC_ADDR)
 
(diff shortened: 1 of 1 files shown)


christophe-lunarg (approved):
This looks good to me. It's something resolved by the `VulkanObject` solution in the validation layr however last time I investigated it many cases of handling alias named were not resolved so it's not ready to use for with the profile solution.

PR #492, opened 2023-08-30
build: Add support for API parameterization

aqnuep:
This introduces similar parameterization of the used API type and layer name as we've done for the validation layers thus I hope it is non-controversial and seen as a general improvement overall.

### DIFF
--- a/layer/CMakeLists.txt
+++ b/layer/CMakeLists.txt
@@ -1,3 +1,7 @@
+# This variable enables downstream users to customize the CMake targets
+# based on the target API variant (e.g. Vulkan SC)
+set(LAYER_NAME "VkLayer_khronos_profiles")
+
 if(BUILD_TESTS)
     add_subdirectory(tests)
 endif()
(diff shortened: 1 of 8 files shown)


layer/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+# This variable enables downstream users to customize the CMake targets
+# based on the target API variant (e.g. Vulkan SC)
+set(LAYER_NAME "VkLayer_khronos_profiles")
```
christophe-lunarg, line 3: I understand this change but...

layer/CMakeLists.txt
```diff
@@ -50,26 +54,27 @@ else()
...
 endif()
 
-source_group("JSON Manifest" FILES VkLayer_khronos_profiles.json.in)
-source_group("Export Files" FILES VkLayer_khronos_profiles.def VkLayer_khronos_profiles.map)
+source_group("JSON Manifest" FILES ${LAYER_NAME}.json.in)
+source_group("Export Files" FILES ${LAYER_NAME}.def ${LAYER_NAME}.map)
 
-add_library(VkLayer_khronos_profiles MODULE)
-set_target_properties(VkLayer_khronos_profiles PROPERTIES FOLDER "Profiles layer")
+add_library(ProfilesLayer MODULE)
```
christophe-lunarg, line 60: why this change?

aqnuep: Please consult with @juan-lunarg.

When doing a similar change to VVL, he recommended that instead of replacing the name of the layer through changing the CMake target, we should use the OUTPUT_NAME to change it.

Technically, the name of the CMake target could be left as is, but it really is historically named as such to match the output binary name, which is no longer relevant, hence a more generic name is probably preferred, just like on the VVL side the target name was changed from VkLayer_khronos_validation to simply vvl.

If there is a problem with that, I can change it, but I just tried to follow the same principle @juan-lunarg suggested on the VVL side.

christophe-lunarg: Ok, I am not seeing the point of that, the name of the binary is a pretty clear name for me for a target but ok.

christophe-lunarg (approved):

juan-lunarg (commented):
LGTM 👍🏾

PR #495, opened 2023-09-07
cmake: Use common LunarG build configuration

christophe-lunarg:
Same work as https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/98 to test if the approach works.


layer/CMakeLists.txt
```diff
@@ -47,10 +48,9 @@ endif()
 add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
 
 if (WIN32)
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
-    set (CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+    set (CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} /bigobj")
 else()
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wpointer-arith")
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wpointer-arith")
```
christophe-lunarg, line 54: @juan-lunarg  Not sure how to handle CMAKE_C_FLAGS , considering that `pointer-arith` is in `lunarg_target_compiler_configurations`

scripts/known_good.json
```diff
@@ -64,7 +64,7 @@
             "sub_dir": "valijson",
             "build_dir": "valijson/build",
             "install_dir": "valijson/build/install",
-            "commit": "v1.0"
+            "commit": "v1.0.1"
```
juan-lunarg, line 67: Please also update the version of valijson in VulkanTools. I'd like them to be consistent to avoid any confusion when we build the SDK.

juan-lunarg (approved):

layer/CMakeLists.txt
christophe-lunarg: Ok, I am revoming it then!

scripts/known_good.json
christophe-lunarg: Ok!

layer/CMakeLists.txt
```diff
@@ -46,22 +47,16 @@ endif()
 
 add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
 
-if (WIN32)
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_CRT_SECURE_NO_WARNINGS /bigobj")
```
juan-lunarg, line 50: FWIW /bigobj looks like it was carried over from VVL. I'd actually like to fix it properly in VVL FWIW because when CPP files get too big it can cause problems/slowdowns in IDEs. Especially when running coding analysis tools.

juan-lunarg (approved):

layer/CMakeLists.txt
christophe-lunarg: Ok I can remove it.

This said, it may be useful because we generated vulkan_profiles.hpp that is 2 MB but here the file remains quite small considering that it's generated using few Vulkan Profiles. But a user could include a lot more profiles so we may hit this case.

CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.17.2)
 
 project(VULKAN_PROFILES LANGUAGES CXX C)
 
+# Enable GUI folders
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
```
juan-lunarg, line 6: FWIW when the cmake min is 3.26 this won't be needed. I added a policy to CMake to make it the default in the future.

https://cmake.org/cmake/help/latest/policy/CMP0143.html

PR #509, opened 2023-09-29
profiles: Update to header 266

ziga-lunarg:


ziga-lunarg:
Run # 2792 was successful on CI, but for some reason message was not reported here, merging.

PR #540, opened 2023-11-08
Fix logging on Android to correctly print formatted input values

kocdemir:
Fixes Android logs to print the input values given, rather than printing the formatting specifiers (e.g. '%s') in the message.


christophe-lunarg:
Thanks for contributing!

PR #544, opened 2023-11-09
profiles: Add Zink feature level profiles

nwnk:
This is sourced from upstream Mesa:

    https://gitlab.freedesktop.org/mesa/mesa/-/raw/main/src/gallium/drivers/zink/VP_ZINK_requirements.json

### DIFF
--- a/profiles/CMakeLists.txt
+++ b/profiles/CMakeLists.txt
@@ -174,6 +174,7 @@ install(
         VP_LUNARG_desktop_baseline_2022.json
         VP_LUNARG_desktop_baseline_2023.json
         VP_LUNARG_desktop_portability_2022.json
+        VP_ZINK_requirements.json
     DESTINATION
         ${json_install_dir}
 )
--- a/profiles/VP_ZINK_requirements.json
+++ b/profiles/VP_ZINK_requirements.json
@@ -1,0 +1,946 @@
+{
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.1-251.json",
+    "capabilities": {
+        "vulkan10requirements": {
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "robustBufferAccess": true
+                }
+            }
+        },
+        "gl21_baseline": {
+            "extensions": {
+                "VK_KHR_maintenance1": 1,
+                "VK_KHR_create_renderpass2": 1,
+                "VK_KHR_imageless_framebuffer": 1,
+                "VK_KHR_timeline_semaphore": 1,
+                "VK_EXT_custom_border_color": 1,
+                "VK_EXT_line_rasterization": 1,
+                "VK_KHR_swapchain_mutable_format": 1,
+                "VK_EXT_border_color_swizzle": 1,
+                "VK_KHR_descriptor_update_template": 1
+            },
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "logicOp": true,
+                    "fillModeNonSolid": true,
+                    "alphaToOne": true,
+                    "shaderClipDistance": true
+                },
+                "VkPhysicalDeviceCustomBorderColorFeaturesEXT": {
+                    "customBorderColorWithoutFormat": true
+                },
+                "VkPhysicalDeviceBorderColorSwizzleFeaturesEXT": {
+                    "borderColorSwizzleFromImage": true
+                },
+                "VkPhysicalDeviceLineRasterizationFeaturesEXT": {
+                    "rectangularLines": true
+                }
+            }
+        },
+        "gl21_baseline_vk10": {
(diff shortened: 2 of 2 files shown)


christophe-lunarg (changes requested):
The profiles are evolving as Zink evolve so it doesn't make sense to add it to this repository...

nwnk:
> The profiles are evolving as Zink evolve so it doesn't make sense to add it to this repository...

Forgive me, but I must disagree. I am personally and professionally invested in upstream zink, and I am entirely happy to sync any future updates to its profiles here (and I'm sure there will be some, I want GLES and OpenCL levels as well). This repository's profiles are used by gpuinfo.org to populate its profile tables, and it would be extremely useful for people to be able to correlate `vulkaninfo` dumps with those feature levels.

If profiles can't deal with versioning or changes then maybe profiles need some work. But frankly the only interesting question is what the newest and best zink thinks it needs, and this profile file is the answer, and if the answer changes I promise I'll let y'all know, because that way I let everyone else know, that's why there's a git repo for profiles is so they can change and be updated, right?

Please merge this. I promise I didn't ask for no reason.

christophe-lunarg:
Well, your PR was not motivated.

Adding @SaschaWillems to the conversation.

The Vulkan Profiles have a version mechanisum but the Zink team kept it to version 1 whenever the profiles was changed. It needs to be updated so the GPUinfo.org database remains meaningful. I am not sure GPUinfo.org actually handle profiles versions either.

So just to clarify, your goal is to populate GPUinfo.org database with information about what GPU support what version of Zink?

That's great, but everyone involved needs to be on the same page for this to be done correctly.

Thanks for contributing !

SaschaWillems:
See https://github.com/SaschaWillems/vulkan.gpuinfo.org/issues/81

Not sure if a zink profile would make sense in the database. What's the actual purpose of that profile, and what benefits do developers have from knowing whether a device supports Zink?

christophe-lunarg:
I think there is value and actually I even think the true value of Vulkan Profiles is that type of use case when we resolve the Vulkan application and the drivers compatibility. Can the application runs on device X,Y,Z? With Zink profiles we can answer can OpenGL applications run on Vulkan drivers X,Y,Z? With the Unreal profiles we get can drivers X,Y,Z run Unreal?

Vulkan application deployment is a huge problem to solve. On one side it allows application developer to figure out explictly whether their application has too many Vulkan capabilities requirements and on the other side it helps Vulkan drivers developers to prioritize, implement, optimize Vulkan capabilities/extensions based on actual usages of the extension to enable for their platform higher quality or performance code path in the Vulkan applications.

My concern is more how to do it properly. If things go well, we could see a good number of profiles per application and profile versioning need to be propertly handle by the Vulkan applications for populating GPUinfo database with meaningful data: 

eg: If the VP_ZINK_gl46_optimal version A profile is marked supported for a driver X but then Zink implement a new extension that is added to VP_ZINK_gl46_optimal version B, then driver X no longer support the newest version of Zink that implement VP_ZINK_gl46_optimal version B.

So proper versioning become critical.

nwnk:
> Not sure if a zink profile would make sense in the database. What's the actual purpose of that profile, and what benefits do developers have from knowing whether a device supports Zink?

As an operating system developer, using Mesa as my driver stack, I foresee a time not very far off when zink may be _the_ OpenGL driver for a given platform, because there simply [is only a Vulkan driver](https://docs.mesa3d.org/drivers/powervr.html) and not a native gallium driver.  If gpuinfo.org knew the zink profiles, I could derive the supported GL level from the Vulkan device reports, and I could do so for systems I'd never touched yet.

As a zink developer, I might get a bug report or support request to run on some [arbitrary Vulkan stack](https://github.com/google/swiftshader). If gpuinfo.org knew the zink profiles, I could just go look to see that it doesn't support `VK_KHR_buffer_device_address`, and not need to beg them for a vulkaninfo report. Better, they could find that answer themselves without needing to ask.

As a middleware or engine developer, I might want to keep running on macOS, but also want to take advantage of GL features that postdate 2010. If gpuinfo.org knew the zink profiles, I could deduce whether zink would give me what I needed if layered atop MoltenVK.

Feel free to substitute ANGLE or DXVK or any other translation later in for zink.

Implicit in all this is that gpuinfo.org uses only `KhronosGroup/Vulkan-Profiles` as its source of profiles. It could use other things, but I can't make that decision for them. But anyone else wanting "the set of defined profiles" would want these all in one place, too.

If the objection is that zink hasn't bumped the revision number in the profile yet, well, sure, we were maintaining it as an in-tree detail. I don't have any problem with incrementing the revision number as we change it going forward and continuing to sync those updates here as the canonical source of truth. We already do this for GL extensions, Vulkan profiles are not different. If I cannot add things to the registry then what exactly is it registering?

PR #546, opened 2023-11-10
ci: Add iOS build to CI

juan-lunarg:
closes https://github.com/LunarG/Projects/issues/714


christophe-lunarg:
@juan-lunarg does it mean we want to support the Profiles layer on iOS? @richard-lunarg ?

richard-lunarg:
Yes, we do want to support Profiles on iOS. Do you not see value in this?

christophe-lunarg:
@richard-lunarg  I can see value, it's more whether we have a plan to see the priority and the focus I should consider on this or not.

juan-lunarg:
> @richard-lunarg I can see value, it's more whether we have a plan to see the priority and the focus I should consider on this or not.

For now the plan is that MacOS testing should be enough to cover iOS testing. So it shouldn't require much if any work from profiles.

PR #582, opened 2023-12-13
build: Make jsonschema module installation optional

christophe-lunarg:


richard-lunarg:
I'm still getting the missing jsonschema module on macOS when I try to build, even with this change.
@christophe-lunarg

PR #583, opened 2023-12-14
build: Fix jsonschema is optional

christophe-lunarg:


richard-lunarg (commented):
Yep, I can see where this was the problem!

PR #592, opened 2024-01-10
script: Remove duplicated properties with merge script

christophe-lunarg:

PR #593, opened 2024-01-10
layer: Fix warning that should be a notification

christophe-lunarg:


ziga-lunarg:
Should this really be just a notification? The extension structure will be ignore and I think the user needs to know that.

christophe-lunarg:
> Should this really be just a notification? The extension structure will be ignore and I think the user needs to know that.

Ok then we agree. because that's what this PR does! :D

ziga-lunarg:
I think you've misread my comment, first part was a question. I think it should be a warning, because the structure will be ignore and the user might not know that.

christophe-lunarg:
@ziga-lunarg, then what was your issue about?

According to the spec, both structure should have the same value. So if it's ignored, fine, it should be the same values. If it's not the same value, then there is a problem in the profile definision.

ziga-lunarg:
Why would a profile specify something that will be ignored anyway? I've recently made changes to vulkaninfo profile generation, so that it doesn't do that anymore. For me it makes more sense for this to be a warning, but if you think notification is enough I will accept it.

christophe-lunarg:
Initially I thought the same and created https://github.com/KhronosGroup/Vulkan-Profiles/pull/592
But vulkancapsviewer is working differently and output everything. And I think it's more consistent with the spec, even if the spec is weird on that topic : Yes it provides both the bundle and independent structures and that weird but that's how it works and the spec only says the value should be the same. A user may want to query a provide with either method so I should say that the change in vulkaninfo is a bug as it requires the user to conform to the method we prefered, a method which I don't even know what it is or know where to find what it is expect by trying, but I should try only because you just tall me otherwise I would just not understand why I am not getting the information I want potentially...

ziga-lunarg:
@christophe-lunarg I think it's more confusing if you have both, promoted and extension structures, because then if you edit extension structures it wont do anything. Sure the spec says you can query both, but a profile will set both of those just from the promoted structure.

christophe-lunarg:
Ok with this case but I don't think that works with all the cases. For example, a user could search for a strcture and not find it and think it's not supported.

So, I am find with reverting this PR and what I can do is to create an option to strip the duplicated structure with the merge script. I think with Vulkaninfo it should also be an option to strip the duplicated structures.

With this approach, I think we get the benefit of your scenario and we document with the option that case. I can generate the Desktop Baseline 2024 profile using that option.

What do you think?

ziga-lunarg:
That sounds good to me

PR #597, opened 2024-01-25
profiles: Update to header 276

christophe-lunarg:


scripts/known_good.json
```diff
@@ -6,7 +6,7 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.275"
+            "commit": "v1.3.276"
```
ziga-lunarg, line 9: I usually update Vulkan-Utility-Libraries and Vulkan-Loader header version as well

ziga-lunarg (approved):

scripts/known_good.json
christophe-lunarg: I'll wait for the repository to be updated and update both Vulkan-Utility-Libraries and Vulkan-Loader

ziga-lunarg:
Build #3110  passed on CI, but no message here

PR #672, opened 2024-04-27
layer: Fix checking device supported extensions

ziga-lunarg:
Fix checking which extension the device supports. Currently only the original extension was checked and not any of the promoted to extensions. This was an issue with a corner case when a device would support a promoted extension but not the original one.

For example a device could support `VK_EXT_mutable_descriptor_type`, but not `VK_VALVE_mutable_descriptor_type`.


christophe-lunarg (approved):

christophe-lunarg:
Nice catch !

PR #691, opened 2024-08-06
Add Android 16 profile

ianelliottus:
This is the "Vulkan Profile for Android 16" (a.k.a. VPA16).  It will be required for chipsets starting or restarting Google Requirements Freeze with Android 16.  It depends on VPA15, which in turn depends on ABP 2022.


christophe-lunarg:
This PR was replaced by https://github.com/KhronosGroup/Vulkan-Profiles/pull/692 to add test and rebase with main to fix the CI failure.

So closing!

PR #694, opened 2024-08-13
WIP: Test valijson changes

mikes-lunarg:
This is a test, do not merge


mikes-lunarg:
So this does work, but a better option is just to use valijson 1.0.2 instead

PR #714, opened 2024-10-28
Default initialize layer inserted VkApplicationInfos

JamesRumble-IMG:
If a VkApplicationInfo is not provided by an application and an apiVersion is requested then the layer inserts its own VkApplicationInfo. This structure should be default initialized to avoid issues for further layers and icds reading this member.


christophe-lunarg:
Thanks for contributing!

PR #729, opened 2024-12-07
layer: Initialize PhysicalDeviceData members

aqnuep:


ziga-lunarg (approved):
LGTM, but was this causing issues?

aqnuep:
No, nothing that I could find, but as the TSAN tests on the VVL showing all sorts of weird slowdowns recently, I'm trying to find out if any of it actually comes from the profile. No luck so far, but getting rid of uninitialized fields should be a good thing nonetheless.

aqnuep:
I just realized I left this PR in draft state for no good reason.

PR #736, opened 2025-01-14
Fix limittype for minMemoryMapAlignment

gnl21:
This was corrected in the official XML v1.3.304 but, because the limittype is overwritten by the profiles generator script as part of a workaround, this was not picked up. Remove the workaround, since the known_good headers already include the fix.

### DIFF
--- a/scripts/gen_profiles_solution.py
+++ b/scripts/gen_profiles_solution.py
@@ -4207,6 +4207,9 @@ class VulkanRegistry():
         if self.headerVersionNumber.patch < 207: # vk.xml declares maxColorAttachments with 'bitmask' limittype before header 207
             self.structs['VkPhysicalDeviceLimits'].members['maxColorAttachments'].limittype = 'max'
 
+        if self.headerVersionNumber.patch < 304: # vk.xml declares minMemoryMapAlignment with 'min,pot' limittype before header 304 (and 'noauto' before 219)
+            self.structs['VkPhysicalDeviceLimits'].members['minMemoryMapAlignment'].limittype = 'max,pot'
+
         # TODO: We currently have to apply workarounds due to "noauto" limittypes and other bugs related to limittypes in the vk.xml
         # These can only be solved permanently if we make modifications to the registry xml itself
         if 'VkPhysicalDeviceLimits' in self.structs:
@@ -4215,7 +4218,6 @@ class VulkanRegistry():
             self.structs['VkPhysicalDeviceLimits'].members['mipmapPrecisionBits'].limittype = 'bits'
             self.structs['VkPhysicalDeviceLimits'].members['viewportSubPixelBits'].limittype = 'bits'
             self.structs['VkPhysicalDeviceLimits'].members['subPixelInterpolationOffsetBits'].limittype = 'bits'
-            self.structs['VkPhysicalDeviceLimits'].members['minMemoryMapAlignment'].limittype = 'min,pot'
             self.structs['VkPhysicalDeviceLimits'].members['minTexelBufferOffsetAlignment'].limittype = 'min,pot'
             self.structs['VkPhysicalDeviceLimits'].members['minUniformBufferOffsetAlignment'].limittype = 'min,pot'
             self.structs['VkPhysicalDeviceLimits'].members['minStorageBufferOffsetAlignment'].limittype = 'min,pot'


gnl21:
This change only fixes the limit that is currently biting me and should probably be updated before actually merging. It looks to me as though a number of the other limits in the workaround are correct in the XML now, so they could be updated consistently. 

Also, I've just removed the workaround since there's a known_good version of the XML that profiles builds against and it has the fix, but other workarounds are conditional on old versions of the XML. Which versions are supposed to be supported here?

gnl21:
@ziga-lunarg, any suggestions for what would be the right fix here? I'm keen to get this resolved because it is currently causing all of the "VP_LUNARG_minimum_requirements_1_x" profiles to report unsupported on many devices.

ziga-lunarg:
> @ziga-lunarg, any suggestions for what would be the right fix here? I'm keen to get this resolved because it is currently causing all of the "VP_LUNARG_minimum_requirements_1_x" profiles to report unsupported on many devices.

I believe removing these is the correct fix, since the vulkan headers now have the correct values. But to clarify, the `limittype` in xml is now different to what the profiles generator script is overwriting it to? And that is why it is failing?

ziga-lunarg:
@gnl21 I see that for some of these overwrites we already have checks, for example:
`if self.headerVersionNumber.patch < 213:`

This seems better than just removing it

gnl21:
Correct. Until recently the XML contained the same as the generator was writing. This was corrected upstream but the overwrite prevents the profiles library from seeing the fix. I imagine that many of the other limits are fixed in the XML but still match what is being written in the script.

gnl21:
> I see that for some of these overwrites we already have checks

This was what gave me pause but since the external dependencies fetch a particular version of the XML (as part of Vulkan-Headers) does that really make sense? There are a variety of old versions needed and I can go through and find them, but I don't want to do that if it's never going to be used.

ziga-lunarg:
I guess a simpler fix that works for all xml versions would be to just check if the properties are `None` and only then overwrite them. It would be good to also changes the wrong overwrites to what xml sets it to now.

gnl21:
I think that the problem is that the XML did have limittypes for those items, the library just considers them to have been wrong, so they won't show up as `None`.

gnl21:
I've updated this to be version dependent like the other one. I can fix the others like this, but I'm concerned that they won't ever really be tested, so if a change to the XML like this gets made in the future then the overwrites will just stick around being wrong.

scripts/gen_profiles_solution.py
```diff
@@ -4207,6 +4207,9 @@ def applyWorkarounds(self):
         if self.headerVersionNumber.patch < 207: # vk.xml declares maxColorAttachments with 'bitmask' limittype before header 207
             self.structs['VkPhysicalDeviceLimits'].members['maxColorAttachments'].limittype = 'max'
 
+        if self.headerVersionNumber.patch < 304: # vk.xml declares minMemoryMapAlignment with 'min,pot' limittype before header 304 (and 'noauto' before 219)
```
ziga-lunarg, line 4210: I think this should still be inside of `if 'VkPhysicalDeviceLimits' in self.structs:`

gnl21:
> If I understand it correctly now, xml had one limittype, profiles overwrote it with another, but now xml limittype was changed to a third value that is different from the original xml and the one in the profiles script?

The full history is that XML contained one value ("noauto") and the script overwrote it with another value ("min,pot"). XML got update and used that second value for a while but it was wrong, so the XML got fixed to say "max,pot" but the generator script was still overwriting it.

scripts/gen_profiles_solution.py
gnl21: Should the same fix be applied to the previous check in 4207? If that condition is not true, is there any chance of generating a useful profile anyway?

ziga-lunarg: Yes I believe so. If `VkPhysicalDeviceLimits` is not specified it is possible to just use device values

gnl21:
Well, yes, but then we run the risk of having the same issue in the future. Leaving the XML information duplicated around in a place where it won't pick up any fixes is not great.

ziga-lunarg:
I guess it would have been better to only overwrite from a known incorrect value, otherwise emit a warning.

gnl21:
How about I add a check that the version is at least 304, otherwise emit a warning asking people to run the update_deps thing to get new XML, then remove all the old workarounds?

ziga-lunarg:
That works for me. @christophe-lunarg any thoughts?

ziga-lunarg:
@gnl21 We agreed internally that we prefer the code as is, overwriting invalid values, rather than emitting a warning. If you would prefer I can take over this PR and add `if` statements for all the limittypes that have been fixed in the xml

gnl21:
Yes please. If you're happy to go through adding the if statements then I'm fine with you taking over the PR. Out of interest, do you plan to test the code with older XMLs to ensure that the overwrites don't just start rotting?

ziga-lunarg:
Wouldn't I have to test with newer XMLs as they come to test the overwrites? Or did you mean something else?

gnl21:
Maybe I've misunderstood what you meant about adding `if` statements, but the `if` statement that is already in the generator script hasn't actually been overwriting anything since 1.3.207, so unless you are running tests with that old XML then the code in the `if` is not being tested.

ziga-lunarg:
I understand now what you mean, I am not sure if this is something we can get into CI, but I will test it manually now when I add these `if`s.

gnl21:
I was really checking that you were sure that you want to add these `if`s rather than requesting more CI tests. I don't really understand the usecase for them and even if someone does end up using them on some system they won't be being tested so they might not work. (The old versions of the XML aren't changing and this code probably won't change either, so the risk is probably low, but if someone refactors something then they probably won't run tests on old versions, so problems can still creep in).

But anyway, it's your code and you've offered to fix my problem for me, so it's all good.

ziga-lunarg:
I see your point, I will try to add CI tests with an older header version, that would cover these paths.

And I think rather than looking for the header version, we should instead check for the known invalid value. That way, when the value gets corrected in the xml profiles will stop overwriting it. Do you see any issues with this?

ziga-lunarg:
@gnl21 Can you confirm #738 fixed your issue and this can be closed?

PR #738, opened 2025-01-19
profiles: Update overwriting xml limittypes

ziga-lunarg:
Replaces #736

### DIFF
--- a/scripts/gen_profiles_solution.py
+++ b/scripts/gen_profiles_solution.py
@@ -4203,259 +4203,221 @@ class VulkanRegistry():
 
         return profileName
 
+    def overwrite(self, structName, memberName, invalid_value, correct_value):
+        if structName in self.structs:
+            if (self.structs[structName].members[memberName].limittype == None or
+                self.structs[structName].members[memberName].limittype == invalid_value):
+                self.structs[structName].members[memberName].limittype = correct_value
+            elif (self.structs[structName].members[memberName].limittype != correct_value):
+                Log.w("Profiles is overwriting {0}::{1} from {2} to {3}, but current XML value is {4}".format(structName, memberName, invalid_value, correct_value, self.structs[structName].members[memberName].limittype))
+
+
     def applyWorkarounds(self):
         if self.headerVersionNumber.patch < 207: # vk.xml declares maxColorAttachments with 'bitmask' limittype before header 207
             self.structs['VkPhysicalDeviceLimits'].members['maxColorAttachments'].limittype = 'max'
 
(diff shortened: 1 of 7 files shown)


christophe-lunarg (approved):

scripts/gen_profiles_solution.py
```diff
@@ -4203,259 +4203,221 @@ def genAliasValues(flagBitsTypeName, map):
...
+        self.overwrite('VkPhysicalDeviceLimits', 'minStorageBufferOffsetAlignment', 'noauto', 'min,pot')
+        self.overwrite('VkPhysicalDeviceLimits', 'optimalBufferCopyOffsetAlignment', 'noauto', 'min,pot')
+        self.overwrite('VkPhysicalDeviceLimits', 'optimalBufferCopyRowPitchAlignment', 'noauto', 'min,pot')
+        self.overwrite('VkPhysicalDeviceLimits', 'nonCoherentAtomSize', 'noauto', 'min,pot')
+        self.overwrite('VkPhysicalDeviceLimits', 'timestampPeriod', 'noauto', 'min,mul')
+        self.overwrite('VkPhysicalDeviceLimits', 'bufferImageGranularity', 'noauto', 'min,mul')
+        self.overwrite('VkPhysicalDeviceLimits', 'pointSizeGranularity', 'max', 'min,mul')
+        self.overwrite('VkPhysicalDeviceLimits', 'lineWidthGranularity', 'max', 'min,mul')
+        self.overwrite('VkPhysicalDeviceLimits', 'strictLines', 'noauto', 'bitmask')
+        self.overwrite('VkPhysicalDeviceLimits', 'standardSampleLocations', 'noauto', 'bitmask')
```
gnl21, line 4238: It looks like this line should say "exact", rather than "bitmask"? That's what the old overwrite used to do and that version seems correct to me.

ziga-lunarg: Original value of `standardSampleLocations` was `noauto` and was changed to `bitmask` by https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5172

Should this be fixed in the spec?

gnl21: Yes, I think this should be fixed in the XML. This looks like a copy-and-paste error in that MR which probably couldn't be tested because the values were all being overwritten by these overwrites. I can make a Vulkan-Docs MR to fix it.

gnl21: I assume that means that we will need 2 overwrites here, one for each of the incorrect values in versions of the XML.

gnl21: The same seems to be true for `strictLines` but I guess that isn't required in any profiles so it didn't show up in my test.

Issue #747, opened 2025-03-05
Update jsoncpp to version 1.9.6


jpr42:
Compatibility with CMake < 3.5 has been removed from CMake 4.0

jsoncpp before version 1.9.6 will cause an error as a result

Link to release notes:
https://github.com/open-source-parsers/jsoncpp/releases/tag/1.9.6

See release notes:
https://cmake.org/cmake/help/latest/release/4.0.html

Also 1.9.6 has a fair amount of bug fixes.

jpr42:
See https://github.com/LunarG/VulkanTools/pull/2297 for reference. Since it's only a patch version update it shouldn't require much change. But it's good to check the release notes in case of a quality of life fix.

PR #760, opened 2025-03-28
profiles: Set json cmake policy version minimum

ziga-lunarg:


jpr42:
Why is this needed? jsoncpp v1.9.6 has a CMake min of 3.8

jpr42:
https://github.com/KhronosGroup/Vulkan-Profiles/issues/747

PR #811, opened 2025-09-15
Add logging tag to android logging

tomnomgoogleuk:
Add logcat tag "vkprofiles". it's convenient to do adb logcat -s vkprofiles to see all vulkan profile related logs.


ziga-lunarg:
@tomnomgoogleuk Thanks for the PR. Can you help me understand what the difference is between the new `vkprofiles` tag and the `Profiles` tag that already exists?

PR #893, opened 2026-06-17
build: add ENABLE_SO_VERSIONING option to CMake

### LINKED ISSUE #892: Add option to enable SO versioning for the library
This allows downstream packaging to not use a dedicated patch. SO versions are usually required in Debian based distributions and track the ABI of said library. Installing via a package manager ensures compatibility with other system packages.

aglinserer:
Introduce a new `SO_VERSIONING` CMake switch to manage shared object versioning. Shared library versioning is required for Debian-based package managers to track ABI breakages properly.

This option is disabled by default to maintain existing behavior, but allows downstream Linux packagers to enable it easily without needing to maintain distribution-specific patches.

This addresses https://github.com/KhronosGroup/Vulkan-Profiles/issues/892

### DIFF
--- a/layer/CMakeLists.txt
+++ b/layer/CMakeLists.txt
@@ -68,6 +68,8 @@ if(BUILD_MOCK_ANDROID_SUPPORT)
     add_definitions(-DVK_USE_PLATFORM_ANDROID_KHR)
 endif()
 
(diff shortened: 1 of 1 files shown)


christophe-lunarg:
Thanks for contributing!

PR #926, opened 2026-08-03
Implement `VK_NO_PROTOTYPES` support #734

### LINKED ISSUE #734: vulkan_profiles.hpp does not compile when VK_NO_PROTOTYPES is defined

christophe-lunarg:
This PR is based on the work provided by https://github.com/KhronosGroup/Vulkan-Profiles/pull/924.

It refactors the VP_USE_OBJECT code path which is designed to replace the Profiles API library design where `VP_USE_OBJECT ` is not defined, a code path that will be deprecated and remove to simplify the code.

This code path enabled using the Vulkan Profiles API with a Vulkan API dynamically loaded


scripts/gen_profiles_solution.py
```diff
@@ -1667,14 +1686,16 @@ def stripNonmatchingAPIs(tree, apiName, actuallyDelete = True):
...
-            capabilities,
+            functions,
 #endif//VP_USE_OBJECT
             &blocks[0].profiles);
     }
 
     VkInstanceCreateInfo createInfo = *pCreateInfo->pCreateInfo;
     createInfo.pApplicationInfo = &appInfo;
 
+    bool use_gpdp2 = false;
```
GrinlexGH, line 1782: You doesn't need that flag, gpdp2 is always enabled by your logic

scripts/gen_profiles_solution.py
```diff
@@ -1152,204 +1227,149 @@ def stripNonmatchingAPIs(tree, apiName, actuallyDelete = True):
...
+#undef VP_COPY_IF_NOT_NULL
 
-        this->CreateInstance = (PFN_vkCreateInstance)vkCreateInstance;
-        this->CreateDevice = (PFN_vkCreateDevice)vkCreateDevice;
-    }
-
-    void ImportVulkanFunctions_Custom(VpVulkanFunctions* pFunctions) {
-    #define VP_COPY_IF_NOT_NULL(funcName) \
-        if(pFunctions->funcName != nullptr) this->funcName = pFunctions->funcName;
+    return functions->validate();
```
GrinlexGH, line 1291: validation is not needed here, because with static initialization by default, all your functions are guaranteed not to be null, and with dynamic loading the user will use a different function

scripts/gen_profiles_solution.py
```diff
@@ -210,66 +210,141 @@ def stripNonmatchingAPIs(tree, apiName, actuallyDelete = True):
...
+typedef VkFlags VpFunctionsCreateFlags;
 
-typedef struct VpVulkanFunctions {
-    /// Required when using VP_DYNAMIC_VULKAN_FUNCTIONS.
-    PFN_vkGetInstanceProcAddr GetInstanceProcAddr;
-    /// Required when using VP_DYNAMIC_VULKAN_FUNCTIONS.
-    PFN_vkGetDeviceProcAddr GetDeviceProcAddr;
+typedef enum VpInstanceFunctionsLoadFlagBits {
+    VP_INSTANCE_FUNCTIONS_LOAD_KHR_GET_PHYSICAL_DEVICE_PROPERTIES2_BIT = (1 << 0),
+    VP_INSTANCE_FUNCTIONS_LOAD_IF_NOT_NULL_BIT = (1 << 1),
```
GrinlexGH: I find this naming a bit confusing, maybe it would be better to rename it to VP_INSTANCE_FUNCTIONS_LOAD_MISSING_ONLY_BIT or something like this? but I could be wrong

GrinlexGH (commented):
This is cleaner and covers a lot of load cases, so it's probably what's needed

scripts/gen_profiles_solution.py
christophe-lunarg: Ok that works for me !

scripts/gen_profiles_solution.py
christophe-lunarg: Actually, I think it is relevant. When using `VK_NO_PROTOTYPES`, the static functions are not initialized and `pFunctionsCreateInfo` should be fill with that other function you are mentioning, the helper function which is `vpInitializeGlobalFunctions`.

If `vpInitializeGlobalFunctions` is not used of `pFunctionsCreateInfo` are not manually filled then `validate` should fail.

scripts/gen_profiles_solution.py
christophe-lunarg: I think it's about time to require Vulkan 1.1 but in this specific code, gpdp2 is only enabled when Vulkan 1.0 is set by the Vulkan application.

christophe-lunarg:
@GrinlexGH the tests are now passing, I only need to update the tutorials and then I think it's ready to merge.

